PR #54, opened 2015-03-02
Update MobRepository to get correct index

Lmctruck30:
The index adds +1 to the player index as players can not have index 0. When you get the index you need to subtract one to get the proper index placement. This will fix this issue.


atomicint:
This can be closed, issue was fixed when I re-based the MobRepository class

PR #55, opened 2015-03-12
Fix and clean graphic synchronization block encoding

Bubletan:
Support for height parameter where missing and cleaning the bit operations.


Major-:
Thanks for the contribution! Apollo already has height support for graphics displayed by players - what you've done is separate it into two shorts, and splitting it isn't such a good idea (although it will work).

NPCs are indeed missing this functionality, and I'm happy to merge this if you update it to remove the changes to player synchronization.

Bubletan:
Alright, reverted the changes made to player synchronization.

PR #60, opened 2015-05-07
Manual if statement in Index.java

basvandriel94:
Why use an library?

### DIFF
--- a/src/org/apollo/fs/Index.java
+++ b/src/org/apollo/fs/Index.java
@@ -17,7 +17,7 @@ public final class Index {
 	 * @throws IllegalArgumentException If the buffer length is invalid.
 	 */
 	public static Index decode(byte[] buffer) {
-		if(buffer.lenght != FileSystemConstants.INDEX_SIZE) {
+		if(buffer.length != FileSystemConstants.INDEX_SIZE) {
 			throw new IllegalArgumentException("Incorrect buffer length.");
 		}
 		int size = (buffer[0] & 0xFF) << 16 | (buffer[1] & 0xFF) << 8 | buffer[2] & 0xFF;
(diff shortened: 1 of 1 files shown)


lare96:
Why did you remove the use of Preconditions#checkArgument for a more verbose if-statement? You also you spelled "length" wrong...

atomicint:
@Major- ping... close...

Hadyn:
Most of the pull requests are like this. Small additions! Gotta get on that contributer list /s

basvandriel94:
It's just one if statement, why use Guava for this?

Hadyn:
They are just simple helper methods to help cut down on verbose code. Either way they achieve the same purpose, I think its just a debate of preference or style. I honestly prefer your method but I see no critical issue writing it in each form. Now in terms of the pull request, if the preference is set in this fashion it's not wise to stray from that so I think it'd be wise to close this request.

MitchZinck:
You're just committing to try and boost your ego. It doesn't add any value to the project.

/r/iamverysmart

basvandriel94:
@Hadyn Alright, it was just my preference indeed.

@MitchZinck Wow man, chill out, i'm just here to help.. I'm also intrested in the project, so why not make some suggestions for changes?

atomicint:
As it currently stands that is (unofficially) how the Apollo code spec deals with false conditions throughout the code-base, the reason why we use it is because it's less verbose and nicer to read in one line than your proposed 3 lines.

Major-:
Hi, thanks for your interest. However, there are a number of good reasons for keeping Guava:
* Shorter code (as is evident from the diff of your commit).
* Reduce repeated code (your replacement would require repeating the if statement and the `new IllegalArgumentException` everywhere, with Guava we are only repeating the method call).
* It's not just used for Preconditions, but for lots of other utilities.

Also, you should check that the codebase compiles and runs before you commit something, as your original commit containing `buffer.lenght` would not have.

Issue #70, opened 2015-08-24
Height level out of bounds issue


WizardJesse1:
When you try to go to height level 4 it will not let you go and there are a few castles like rouges castle in the wilderness that does need this fixed if you want to go all the way too the top.

Major-:
When you say "height level 4", do you actually mean 3? In apollo the height levels are 0-3 inclusive, there is no 4.

WizardJesse1:
ya that is what i meant @Major-

ryleykimmel:
So what's the issue? You need to go to 4 or you need to go to 3? IIRC the only bounds issue that would occur is when you meet or exceed height 4, not 3.

ryleykimmel:
Never mind -- I see the flaw with the check, making a pull request in a second.

ryleykimmel:
@Major- close

PR #71, opened 2015-08-25
Fixes issue #70

ryleykimmel:
Fixes the check in the Position constructor, also includes a simple test.


Major-:
Nice (particularly for the unit test), thanks.

PR #73, opened 2015-08-27
Update dependencies, do not fail startup when we cannot bind to HTTP (just use jaggrab as a fallback)

ryleykimmel:


Major-:
Cheers (in future check out 'git rebase' instead of merging commits from master into a local branch).

Issue #75, opened 2015-08-28
Various Animation instances should be moved to the emote_tab plugin.


Major-:
The `Animation` class contains instances of various Animations like so:

```java
/**
 * The angry animation.
 */
public static final Animation ANGRY = new Animation(859);
```

These are only used in the emote_tab plugin and should be removed from the core server.

PR #83, opened 2015-08-29
Moved Animation constants from Animation.class to emote_tab.rb

lare96:
Didn't include STOP_ANIMATION (obviously)


Major-:
Great, thanks. This fixes issue #75.

PR #85, opened 2015-08-29
New players will be queued rather than added right away.

lare96:
Let me know if anything needs to be changed :)


Major-:
Can you reformat it to use tabs instead of spaces, and also squash those two commits into one.

lare96:
Not sure what to do past here http://prntscr.com/8ad8pz

Major-:
@lare96 hit Esc to exit insert mode, then :wq to save and quit

lare96:
okay lol it seems as if it didn't remove the previous commits

PR #90, opened 2015-08-30
Preconditions instead of if-statements for the GamePacketBuilder class.

lare96:


Major-:
Can you make [this check](https://github.com/lare96/apollo/commit/b8223e5787d022a25fb8b7f78717c99891886fe0#diff-be5d6183b709ad4765381296d8c749c3R226) >0 not >= 0; looks like there was a bug in the previous implementation of it, but it should be [1, 32].

lare96:
Done

Major-:
Great, thanks. Please look into git rebase and squashing, which will let you turn several tiny related commits into one. Not an overly big deal, but it's a lot nicer when looking through the commit history.

Issue #76, opened 2015-08-28
Add ban/mute support


Major-:
Add support for banning/muting users.

Major-:
@lare96 @ryleykimmel would one of you be interested in this?

Promises:
Should this be done via sql or?

Major-:
@Promises apollo supports [several methods of serialization](https://github.com/apollo-rsps/apollo/tree/master/game/src/main/org/apollo/game/io/player), so it would depend on which one you are using. Right now the SQL one is non-existent, so it would only be added for the binary serializer.

Ban/mute status should be saved with the rest of the stuff for each player, not separately.

lare96:
@Major- I wouldn't mind, would both of the flags be manipulated as attributes?

Major-:
Yes, I think that's probably the way to go.

Major-:
Added in c3ebfb4.

PR #95, opened 2015-09-02
Ban and mute support, commands included.

lare96:
To take care of issue #76

### DIFF
--- a/data/plugins/cmd/punishment/plugin.xml
+++ b/data/plugins/cmd/punishment/plugin.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0"?>
+<plugin>
+  <id>punishment</id>
+  <version>1</version>
+  <name>Punishment commands</name>
+  <description>Adds various punishment commands, such as banning or muting a player.</description>
+  <authors>
+    <author>lare96</author>
+  </authors>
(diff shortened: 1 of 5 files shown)


Major-:
Have you tested this thoroughly @lare96?

Major-:
I've made some comments, can you update the code accordingly? If you commit the changes then GitHub should automatically include the new commit in this pull request.

lare96:
`org.jruby.exceptions.RaiseException: (NoMethodError) undefined method 'nil' for # <Java::OrgApolloGameModelEntity::Player:0x15bade79>`

When trying to change on_player == null to on_player.nil

lare96:
Also, I've tested everything but I can't test the muting itself due to issue #96 and for some reason when logging in after being banned, an unexpected response message is given rather than "your account has been disabled." This seems to happen with every response code except the "OK" one. This could be my client though, as I'm using the one you released which is apparently buggy.

Major-:
It needs to be `nil?`, not `nil`

lare96:
Finally figured out how to squash commits correctly, check it over and see if I missed anything

Major-:
Nice job, thanks.

PR #99, opened 2015-09-13
Add simple tests around 'important' Message verification handlers

garyttierney:
## Testing the MessageHandler implementations

This introduces a couple of tests which verify that several of the MessageHandler implementations (Chat, ItemOnItem, ObjectAction) terminate messages appropriately. These few examples outline how tests like this could be written.

One issue I did find was that there is no verification that an object exists (as far as I can tell) in an ItemOnObject message. I'll write a few more tests and update any failing code as needed.

## Notes

Admittedly mocking isn't necessarily needed here because we aren't accessing the file system or any system resources, but setting up all of these classes with all of their dependencies would be mayhem. 

## New Dependencies

* Mockito - A library for mocking classes in tests
* PowerMock - A mocking library to work around final classes and static methods

### DIFF
--- a/game/src/main/org/apollo/game/message/handler/ItemOnObjectVerificationHandler.java
+++ b/game/src/main/org/apollo/game/message/handler/ItemOnObjectVerificationHandler.java
@@ -1,12 +1,19 @@
 package org.apollo.game.message.handler;
 
+import org.apollo.cache.def.ObjectDefinition;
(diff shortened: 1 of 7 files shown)


Major-:
Is this ready to be merged or are you still updating it?

garyttierney:
@Major- ready to merge

PR #100, opened 2015-09-19
Adds diagonal movement to SimplePathfindingAlgorithm class

DrowsySaturn:
Adds diagonal movement to the SimplePathfindingAlgorithm class


DrowsySaturn:
Needs a fix.

Issue #94, opened 2015-09-02
Bank Noting/Unoting


WizardJesse1:
Doesn't work and I always though it did for some reason.

WizardJesse1:
This can be closed

PR #108, opened 2015-12-21
Fix short circuit logic in ItemDefinitionDecoder

garyttierney:
* Replaces an OR, with what should have been an AND while decoding
  unused item definition info.


garyttierney:
Fixes #94 which is closed for some reason? Possibly unrelated.

PR #110, opened 2015-12-22
Correct the height levels of steps in a path

### LINKED ISSUE #109: Npc Height issue
Pretty much if you go +1 height u see all the npcs from height 0

http://puu.sh/m3K2r/8a4905ab79.jpg

MORE:

When a player walks on height 1 it seems like it is setting the players height to 0 which is why this is happening

garyttierney:
Makes sure that the steps in a path sent for a "Player Walk" message are
on the same height level as the player that sent them beforethey are
added to the walking queue.

Fixes #109.

This is a bit of a crude fix, I think. Could this be better done in whatever handles steps in the WalkingQueue? cc @Major-


garyttierney:
See new commits for a proper fix

Issue #107, opened 2015-12-13
Login issue after log out


Lmctruck30:
Just downloaded a fresh copy and when you log in then log out you will get a stuck log in screen with no errors other then the client saying error connecting to server.

Mac running latest version of java 8.66. Client is a non edited MITB client.

WizardJesse1:
I know this issue but I don't believe its the best way to fix

WizardJesse1:
I have fixed it though

lare96:
@WizardJesse1 share the fix

WizardJesse1:
@lare96 i would love to share all my fixes if i knew how to setup github rofl

lare96:
pastebin it if you have to

WizardJesse1:
@lare96 I have been going to @garyttierney and telling him the bugs so i will give him the fixes if he wants to push them for me lolol

WizardJesse1:
@Lmctruck30 http://recordit.co/QU9JhihGXD this?

packersfan84:
epic nice

ryleykimmel:
This is still an issue, @Major- am able to replicate it.

Major-:
Rats

Major-:
cc @garyttierney is this fixed in your local copy?

garyttierney:
@Major- was never able to reproduce

Major-:
cc @ryleykimmel please look into this as I haven't managed to reproduce it either

ryleykimmel:
I can reproduce this every time. Log in, log out, try to log back in. Using master w/ no edits

Rhys:
I am also able to reproduce this issue. Using master without any edits and running the 377 release.

Artyom-M:
Hi, this also happens for me. Unedited, 377. Jesse's fix seems to work, though.

PR #114, opened 2015-12-29
Fixed Issue #107

WizardJesse1:
Has to do with this issue https://github.com/apollo-rsps/apollo/issues/107


Jire:
Kek Cheap hax since scu/cube complain like little bitches and need it fixed apparently

Major-:
This won't build, and this isn't even a real fix. If this actually solves the problem then the issue is that players are being removed during the update procedure, instead of before/after.

WizardJesse1:
@Major- i will try and look further into this issue when i have time

PR #117, opened 2015-12-29
Fixed Magic unexpected return issue

WizardJesse1:
It was leaving an issue org.jruby.exceptions.RaiseException: (LocalJumpError) unexpected return
	at RUBY.block in (root)(C:\Users\jesse\Desktop\apollo\.\data\plugins\skill\magic\magic.rb:153)
	at org.jruby.RubyProc.call(org/jruby/RubyProc.java:308)
	at RUBY.handle(C:\Users\jesse\Desktop\apollo\.\data\plugins\bootstrap.rb:88)

there was no reason for the returns?

### DIFF
--- a/data/plugins/skill/magic/magic.rb
+++ b/data/plugins/skill/magic/magic.rb
@@ -130,7 +130,6 @@ on :message, :magic_on_item do |player, message|
     item = player.inventory.get(slot)
     player.start_action(AlchemyAction.new(player, alch, slot, item))
     message.terminate
-    return
   end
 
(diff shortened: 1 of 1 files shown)


Major-:
Can you put the enchantment part inside an `else` clause - with this change we're going to check if the spell is an enchantment spell even if we know it's not (because if it's an alchemy spell then it's definitely not an enchantment spell).

garyttierney:
Shouldn't it just be `next`?

PR #119, opened 2015-12-30
Fixed Player Already Online

WizardJesse1:
It use to always finalize the player even if it failed with a status


Major-:
Nice spot, thanks

PR #123, opened 2016-01-20
Game packets encoded with MTB Encoder instead of MTM Encoder.

lare96:
Very small PR that encodes game packets with `MessageToByteEncoder` instead of `MessageToMessageEncoder`. Unless there was a specific reason why Graham used the latter?


lare96:
If this gets accepted/merged then the same thing could be done to `LoginEncoder` as well.

garyttierney:
@lare96 could you include the patch for LoginEncoder in this PR before I merge?

lare96:
@garyttierney done!

garyttierney:
Fixed in https://github.com/apollo-rsps/apollo/commit/7106ba7df798192705b41f3661b4256a49a410c1. Made the commit message a bit neater before merging.

PR #124, opened 2016-01-21
GamePacketEncoder unit test fix for MTM -> MTB encoder.

lare96:
Fixes the GamePacketEncoder unit test, there were errors migrating from the MTM encoder to the MTB encoder.


Major-:
Thanks for this, more unit tests are always appreciated

PR #139, opened 2016-01-31
Fixed small issues with alchemy

WizardJesse1:
* Low and high alch have there own delay now when they use to share the same.

* Runes are checked for the main rune before the elemental which its how its done in runescape.

* Only item that is not allowed to be alched is coins I checked tokkul as well and its allowed.

* Fixed gold amount received it was way off.


Major-:
Argh I forgot to fix the gold amount yesterday. This new one is indeed correct, thanks.

LGTM @garyttierney

PR #147, opened 2016-01-31
Fixed food eating delays

WizardJesse1:
* Removed the equal check due to causing issue's with food delays since when you eat one food it should be on the same tick

### DIFF
--- a/data/plugins/consumables/consumable.rb
+++ b/data/plugins/consumables/consumable.rb
@@ -32,7 +32,7 @@ class ConsumeAction < Action
   attr_reader :consumable
 
   def initialize(player, slot, consumable)
-    super(0, true, player)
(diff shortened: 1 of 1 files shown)


Major-:
I think this is okay, but unfamiliar with this behaviour.

WizardJesse1:
@Major- i can show a video of how it use to be if you want?

Major-:
That would be good (and ideally the new behaviour too), thanks

WizardJesse1:
Before - http://recordit.co/DYPHeiMxfj

After - http://recordit.co/4Lzm8dsEaT

@Major-

Major-:
Ah yeah, you're right, this new behaviour is correct. LGTM cc @garyttierney

PR #151, opened 2016-02-02
separate potion and food consume delays

thispixel:
fixes the bug where you cannot eat and drink a potion at the same time.


garyttierney:
:+1:

Major-:
Good for now, but will need something more nuanced than this because [karambwan](http://runescape.wikia.com/wiki/Cooked_karambwan) runs on the potion delay but is obviously a food

Issue #153, opened 2016-02-03
Change synchronization implementation to iterate over nearby players in surrounding regions


ryleykimmel:
Currently Apollo iterates through _all_ the players. Apollo should only iterate through players in surrounding (viewable) regions.

PR #159, opened 2016-02-04
Changed synchronization implementation to iterate over surrounding regions for nearby mobs

ryleykimmel:
This resolves issue #153


ryleykimmel:
Don't really have a choice, can't keep track of the 'added' amount and break out of the loop if you use forEach

Major-:
Ah yes, sorry.

LGTM

ryleykimmel:
Changed commit message

PR #168, opened 2016-02-07
Fixed Player Option Order

WizardJesse1:
The order for player options were wrong. It's all in the correct order now


Major-:
Great, thanks

PR #193, opened 2016-02-10
Fix Magic Check Issues

WizardJesse1:
I forgot to check a few things when i changed magic check's

Issues Fix

* You couldn't teleport but now you can
* Illegal items would throw an error
* Process elements would get removed even if you couldn't cast the spell due to moving illegal items


WizardJesse1:
@Major- if i left it where it was originally it wasn't working?

PR #197, opened 2016-02-10
Fix Magic Teleporting Delays

WizardJesse1:
* Magic Teleporting delays
* Correct gfx for teleporting
* Teleporting never used a end animation


Major-:
Can we a get a before/after gif?

WizardJesse1:
@Major- http://recordit.co/o2l1P6Awjp

here is a video of everything to the original also before the magic delay pr

Major-:
Excellent, if you could squash this I'll merge it

PR #198, opened 2016-02-10
Fix Teleporting Spam Clicking

WizardJesse1:
You can no longer mass click teleporting


WizardJesse1:
@Major- 

If you ask why i removed `@spell == other.spell` it's due to that it was causing you to override into a different spell while in the middle of another

Major-:
Okay LGTM thanks, please squash and then I'll merge

PR #206, opened 2016-02-11
added test case in methods encodedBases37 

CharlesVaneenoo:
Hello, 
I add a new case in the method test for encodedBases37.
I see the case for a character between 'A' and  'Z' was not tested so i did it.


Major-:
Excellent contribution, thanks

PR #207, opened 2016-02-12
More Npcs to Al-Kharid

WizardJesse1:
* Added every npc that was missing from al-kharid
* Removed and moved any npcs that were not in the correct area


Major-:
Great contribution, thanks

PR #209, opened 2016-02-12
Little change for better understanding

KGALLET:
Hello,

I'm currently implementing test for the EquipmentDefinition class. 
When I read the init method, I considered that it is a bit incomprehensible because the name of the parameter and the HashMap were the same.


garyttierney:
Hi,

I think that the fully qualified access to `EquipmentDefinition.definitions` provides enough clarity, since `definitions` is only available within the scope of that method for a short time. It is also very common to name constructor arguments with the same name as the member they're being assigned to and qualifying the member with `this`, and I think that carries over to this similar static initializer method. 

Thanks.

PR #210, opened 2016-02-12
Implementation of the EquipmentDefinition test class

KGALLET:
Hi,

I implemented test for the EquipmentDefinition class in `/cache/src/main/org/apollo/cache/def/`.  
I don't know if it was usefull to test the getters and setters but just in case, i implemented it.

### DIFF
--- a/cache/src/main/org/apollo/cache/def/EquipmentDefinition.java
+++ b/cache/src/main/org/apollo/cache/def/EquipmentDefinition.java
@@ -62,6 +62,13 @@ public final class EquipmentDefinition {
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


ryleykimmel:
Thanks for the PR and the unit test.

Could you ensure that your code style is consistent with the rest of Apollo as well as address the line-comments in your commit?

garyttierney:
Hi,

The test cases included here only test accessors, and while those tests do prove that the code in question is working correctly we can safely assume that the behavior of these accessors is as expected.

Additionally, please see the above comments from @ryleykimmel on how to improve any code you might want to merge upstream in the future.

Thanks.

PR #211, opened 2016-02-12
remove an useless condition

CharlesVaneenoo:
Hello,
I removed a useless condition.
What do you think about transforming the method with a switch case (for all the case with an equality and "opcode >= 30 && opcode < 40" and "opcode >= 90 && opcode <= 92"  in default case)  ?


garyttierney:
Hey,

Good find. Thanks. 

With regards to refactoring this class to use a switch statement, I think it'd add more boilerplate code than necessary. Handling all the cases from >= 30 and <= 40 is much nicer with the current code than the following:
```java
switch(opcode) {
   case 30:
   case 31:
   case 32:
   case 34:
   case 35:
   case 36
   case 37:
   case 38:
   case 39:
   case 40:
       /* ... */
       break;
}
```

with a huge amount of cases to handle, this quickly can get out of hand as it already is with just those 10! Appreciate the suggestion though, and welcome any more.

PR #212, opened 2016-02-12
add catch exception in EquipmentUpdater

CharlesVaneenoo:
Hello,
We see a try without catch.
We think it's better if we add a catch, because the program will not crash.


garyttierney:
Hi,

Thanks for the contribution. We're happy to be getting lots of pull requests coming in.

I had a look at your patch and I can see why the change might look like it causes the application to exit gracefully in cases where an Exception is passed back down to the `main` method. However, when the JVM is preparing to execute a program it creates a thread to run the main method and sets up the default exception handler for that thread (see the relevant section about uncaught exception handling in the [Java Language Specification](http://docs.oracle.com/javase/specs/jls/se8/html/jls-11.html#jls-11.3)) which prints the exception and terminates execution.

So as this change doesn't change behavior and introduces additional boilerplate I'm closing this.

PR #215, opened 2016-02-13
Fix Several Magic Issues

WizardJesse1:
* Fix OverRide Magic Spells
* Fix Process Element check
* Reordered the rune amounts so now it can be read clearly of how runescape has it by (primary, elemental)
* Changed MAGIC_SKILL_ID for add exp to Skill::Magic in enchant so you now gain exp instead of an error message
* For spells like alching they should always bring you back to your spell book even if you do not have the level requirment or the runes.

### DIFF
--- a/data/plugins/skill/magic/magic.rb
+++ b/data/plugins/skill/magic/magic.rb
@@ -46,7 +46,10 @@ class SpellAction < Action
         stop
         return false
       end
-      process_elements
+      unless process_elements
+        stop
+        return
+      end
     end
     execute_action
     @pulses += 1
@@ -81,7 +84,7 @@ class SpellAction < Action
   end
 
   def equals(other)
-    get_class == other.get_class && @spell == other.spell
+    get_class == other.get_class
   end
 
 end
@@ -100,6 +103,8 @@ class ItemSpellAction < SpellAction
   def execute
     if @pulses == 0
       id = @item.id
+
+      mob.send(DISPLAY_SPELLBOOK)
 
       # TODO: There has to be a better way to do this.
       @spell.elements.each do |element, amount|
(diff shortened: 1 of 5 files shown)


data/plugins/skill/magic/teleport.rb
```diff
@@ -83,14 +80,14 @@ def ancient_tele(*args)
...
 # Modern teleports
-tele 1_164, 25, { FIRE => 1, AIR => 3, LAW => 1 }, 3213, 3424, 35, 'Varrock'
-tele 1_167, 31, { EARTH => 1, AIR => 3, LAW => 1 }, 3222, 3219, 41, 'Lumbridge'
-tele 1_170, 37, { WATER => 1, AIR => 3, LAW => 1 }, 2965, 3379, 47, 'Falador'
-tele 1_174, 45, { AIR => 5, LAW => 1 }, 2757, 3478, 55.5, 'Camelot'
-tele 1_540, 51, { WATER => 2, LAW => 2 }, 2662, 3306, 61, 'Ardougne'
-tele 1_541, 58, { EARTH => 2, LAW => 2 }, 2549, 3114, 68, 'the Watchtower'
-tele 7_455, 61, { FIRE => 2, LAW => 2 }, 2871, 3590, 68, 'Trollheim'
-tele 18_470, 64, { FIRE => 2, WATER => 2, LAW => 2, Element.new([1963], nil, 'Banana') => 1 },
+tele 1_164, 25, { LAW => 1, AIR => 3, FIRE => 1 }, 3213, 3424, 35, 'Varrock'
```
kylestev: reordered the rune requirements :+1:

data/plugins/skill/magic/magic.rb
```diff
@@ -101,6 +104,8 @@ def execute
     if @pulses == 0
       id = @item.id
 
+      mob.send(DISPLAY_SPELLBOOK)
```
kylestev: thank you for pulling this behavior up into the `execute` method. I can't think of any spells that do not return to the spellbook tab after casting :+1:

PR #217, opened 2016-02-13
Fix many issues with the Tutorial island plugin

Major-:
cc @garyttierney @ryleykimmel needs CR


Major-:
Just realised I removed the spawn queueing which we actually need... argh

garyttierney:
:+1:

PR #223, opened 2016-02-14
Fix #221

### LINKED ISSUE #221: Equipment Disappearing
How to produce

http://recordit.co/Nc07i8DLKs

Major-:
cc @garyttierney thoughts on the usage of static imports here pls


garyttierney:
:+1:

I think the usage of static imports here is fine given the limited scope.  Qualifying access to them doesn't provide any additional information, I don't think.

Major-:
rebased

PR #227, opened 2016-02-15
Fix Beards for helmets

WizardJesse1:
* Beards for items such as coifs, med helmets should be fixed now


Major-:
Can you do the 317 one as well? We're still supporting 317 for a bit, but 377 is just going to be the default.

PR #231, opened 2016-02-15
refactor decode methods by removing some redundant if

CharlesVaneenoo:
Hello,
After my PR #211, i think it's possible to improve again the method decode.
So i decide to remove some redundant if to reduce this method.


Major-:
Hi there, thanks for the PR

The reason those are all arranged like that is because each branch decodes different values from the data. Right now we don't know exactly what values we're going to need, so we've left it like this to make it easier to add support for whatever extra values we may need (and this is also true for `ObjectDefinitionDecoder` etc too). Although this change is nice in that it removes a lot of duplicate code, it also makes it more difficult to add support for decoding other values in future, so I'm going to close this.

Thanks again for the PR

PR #232, opened 2016-02-15
add test for Inventory.java

pheq:
Hello,
I added some tests for the inventory class because there weren't any.
Have a nice day.


Major-:
Hi there, thanks a lot for the PR. All of the Inventory stuff is going to go through a big rewrite soon, so I'm going to close this as most or all of these tests will become invalid once this rewrite happens. I'll also try and document the stuff that's going to change soon, so people don't end up having their PRs rejected through no fault of their own - sorry about that.

Thanks again for the PR

PR #234, opened 2016-02-16
Equipment bug fixes

WizardJesse1:
* Fix Several Items that were glitched when worn.
* Fix items that were not able to be worn due to the name.
* Added prayer check in due to one of the items using it during 377


WizardJesse1:
@Major- ??

ryleykimmel:
You removed a space and a tab in your commit.

PR #236, opened 2016-02-16
add a magic number in the Player

AymericDu:
Hello,

I saw a "TODO magic constant" in Player class. So I added a variable for readability.


Major-:
Hi, thanks for the pull request

Check out Wikipedia's [article on magic constants](https://en.wikipedia.org/wiki/Magic_number_%28programming%29#Unnamed_numerical_constants) for what this comment means. We need to know what the constant actually represents - simply calling the field `MAGIC_NUMBER

PR #237, opened 2016-02-16
Adding missing throws statement

munggs:
Hello,

Not a big change here, I added a missing throws statement in the constructor of Appearance.java. I also updated the javadoc in order to match the throws statement.

### DIFF
--- a/game/src/main/org/apollo/game/model/Appearance.java
+++ b/game/src/main/org/apollo/game/model/Appearance.java
@@ -38,8 +38,10 @@ public final class Appearance {
 	 * @param gender The gender.
 	 * @param style The style.
 	 * @param colors The colors.
+	 * @throws NullPointerException if any argument is null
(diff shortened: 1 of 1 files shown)


Major-:
Hi there, thanks for the contribution

It is unusual and unrecommended to declare most unchecked exceptions in a `throws` clause, and both `NullPointerException` and `IllegalArgumentException` are unchecked, so I'm going to close this. You can check out [Should I declare unchecked exceptions in the throws specification?](https://stackoverflow.com/questions/20871763/should-i-declare-unchecked-exceptions-in-the-throws-specification) and [Should methods that throw RuntimeException indicate it in method signature?](https://stackoverflow.com/questions/824217/should-methods-that-throw-runtimeexception-indicate-it-in-method-signature) if you're interested in more information on this.

Thanks again

PR #240, opened 2016-02-17
added comments in dummy.rb

CharlesVaneenoo:
Hello,
i added missing comments in dummy.rb


Major-:
Hi there, thanks for the contribution.

These comments aren't too useful for the reader (this code is pretty simple, so maybe it doesn't need any at all). I recommend checking out [Code Tells You How, Comments Tell You Why](https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/).

CharlesVaneenoo:
Thanks

PR #242, opened 2016-02-19
adding test for Npc.java

munggs:
Hello,

So I wanted to do some tests for Mob.java because I feel like there are some methods worth testing. But with Mob.java being an abstract class I wanted to actually test these methods with Npc.java (which extends Mob.java). I'm not sure whether to put these tests in a MobTest.java or in NpcTest.java.
In both cases it felt weird to use Npc without testing it so here it is.

Let me know if you want me to finish NpcTest.java with methods of Mob.java or if you'd prefer a proper MobTest class

### DIFF
--- a/game/src/test/org/apollo/game/model/entity/NpcTest.java
+++ b/game/src/test/org/apollo/game/model/entity/NpcTest.java
@@ -1,0 +1,59 @@
+package org.apollo.game.model.entity;
+
+import static org.junit.Assert.*;
+import static org.powermock.api.mockito.PowerMockito.mock;
+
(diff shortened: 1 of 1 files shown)


Major-:
Hi there, thanks for contributing

It's unnecessary to test getters or setters, as they're so very simple you aren't gaining anything from having the tests. Unit tests are primarily used for identifying regressions (it used to work, now it doesn't) and for verifying corner-case behaviour - you can see my comment [here](https://github.com/apollo-rsps/apollo/pull/245#issuecomment-189179348) on writing tests that are more useful.

PR #243, opened 2016-02-21
Recipes and Skill Menus

shamonj03:
Review it see how you like it. Good examples are the cooking, herb, and fletching scripts. I don't have much time to finish stuff at the moment, so this is just an alpha release I guess. I guess I uploaded some files that weren't really changed/needed to be deleted. Runecrafting had errors, but this system can be used to rewrite it. Speaking of runecrafting...I guess the rewards need a proc for it as well.

I should explain what this does.
Creation of recipes for item on item (fletching), item on object (cooking/runecrafing), and item option messages (herbs). Recipes are basically requirements and rewards. If requirements met, then assign rewards. Recipes can have multiple materials required to create something and produce a product. They have skill requirements, material requirements, object requirements, and tool requirements. If there happens to be an object requirement it is an item on object event. For single item recipes there is a special main ingredient requirement. For an example of a lot of the features look at the cooking plugin.

Also created a skill menu system that allows you to add items to a skill menu for a specific tab. If that tab is not there it gets created. For recipes the product created is registered to a skill tab. You must explicitly state it's skill and the tab it will be registered to. Also to register it it must be called after the product and skill requirements have been defined.


WizardJesse1:
it looks like you are trying to go back a few updates from looking at shops and etc.. did you not update your source to the current? @shamonj03

shamonj03:
Uh. It's a couple weeks old.

shamonj03:
I don't like my recipe class. It turned into a cluster fuck.

WizardJesse1:
So confused what's going on

shamonj03:
Just delete this

ryleykimmel:
I actually really like the idea you have here.

Major-:
Yeah I think this shows a lot of promise. Do you need a hand with git?

PR #245, opened 2016-02-22
added test methods for readSmart method

CharlesVaneenoo:
Hello,
I noticed that the method readSmart isn't tested, so i added the test method.


Major-:
Hi there,

While you have added a test, this one isn't as useful as it could be. Writing good unit tests takes practice, but I recommend checking out [How to write good unit tests](https://programmers.stackexchange.com/questions/21133/how-to-write-good-unit-tests) for a quick overview of what they should be. The problem with this one is that it doesn't test all possible states of the method - if you look at `readSmart`, you will see that the behaviour is different depending on the initial value it receives (and that will cause it to sometimes read extra data). A good test would cover both of these possibilities

PR #246, opened 2016-02-25
Simplify the getter of Crcs

CharlesVaneenoo:
Hello,
We found a way to simplify the method getCrcs so we did it.
Thanks


Major-:
Nice find, thanks

PR #251, opened 2016-02-26
add new structure for the coordinates in Area

AymericDu:
Hello,

We saw a TODO with @CharlesVaneenoo. So we create a new class inside Areas.rb for the coordinates.
Thanks.

### DIFF
--- a/data/plugins/areas/areas.rb
+++ b/data/plugins/areas/areas.rb
@@ -30,24 +30,24 @@ class Area
     @actions = actions
   end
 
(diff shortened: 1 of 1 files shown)


Major-:
Hi there, thanks for the PR

The idea behind this comment was to implement some sort of quadtree for the areas, although admittedly this isn't specified in the comment. This change just places the data in another class and simply delegates - the quality of the code isn't better than it was before, and it will be (slightly) slower. I'll update the `TODO` comment to clarify what the idea of it is

PR #254, opened 2016-02-26
refactor wrap in dialogue.rb

AymericDu:
Hello,

I refactored the wrap method in dialogue.rb. I think it's easier to read and test the code. I also corrected some bug.

Thank.


Major-:
This is a good start towards improving this code, thanks

PR #259, opened 2016-02-29
refactor the method loadPlayer

CharlesVaneenoo:
Hello,
I decide to refactor the method loadPlayer in order to simplify this one. I think this methods is very long so i split it with a new method setPlayer. 
Thanks.


Major-:
Hi there, thanks for the PR

While the original method is long, I think the new method actually makes this code *less* readeable, because of the very large amount of parameters. I'm not sure if there's much we can do to improve the original method - the vast majority of the time methods should be under 30 lines of code, but sometimes trying to split up a method only makes it harder to read, rather than easier.

Thanks again

PR #262, opened 2016-03-03
improve the performance of the segment_text method in dialogue.rb

AymericDu:
Hello,

I saw that we calculated the size of text for each iteration, which is very costly in computation time. Now I use a variable that contains this size and is refreshed only when we change the text.

Also I deleted the character at the end of the line because it is useless.


cubeee:
Seems a bit unnecessary and `segments << text[0..(space - 1)]` may have unintended results if the message starts with a space

Major-:
"Premature optimization is the root of all evil" - DonaldKnuth

Check out [Program Optimization](https://en.wikipedia.org/wiki/Program_optimization) for information about how and when to optimise code.

PR #263, opened 2016-03-04
Move RSA keys to their own file and use a pem file for them

cubeee:
RSA keys are sensitive information that should not be stored in git repositories.
This commit moves them to their own file, adds it to .gitignore and provides a template file.


garyttierney:
I think storing the key PKCS#8 would be a better option since it's something that NSS, OpenSSL, and the JDK can all work with.

Major-:
Originally this was for ease-of-setup (the public counterparts are in all the clients I made available to people) but in general it would be loads nicer to have a proper ./configure script that generates the keys etc locally

garyttierney:
@cubeee status on this?

cubeee:
@garyttierney done unless you had something in mind

garyttierney:
@cubeee LGTM, just needs rebase.

net/src/main/org/apollo/net/NetworkConstants.java
```diff
@@ -76,6 +74,20 @@
...
+			PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(pem.getContent());
+
+			Security.addProvider(new BouncyCastleProvider());
+			KeyFactory factory = KeyFactory.getInstance("RSA", "BC");
+
+			RSAPrivateKey privateKey = (RSAPrivateKey) factory.generatePrivate(keySpec);
+			RSA_MODULUS = privateKey.getModulus();
+			RSA_EXPONENT = privateKey.getPrivateExponent();
+		} catch (Exception exception) {
+			throw new ExceptionInInitializerError(new IOException("Error parsing id_rsa", exception));
```
garyttierney: Confusing error message, id_rsa doesn't exist

PR #267, opened 2016-03-05
*Fix issue #166

WizardJesse1:
* Currently updated to how it should be.


WizardJesse1:
and it seems like i messed up...

PR #270, opened 2016-03-07
Add support for projectiles

garyttierney:
* Adds support for short-lived entities, which are never added to the
  entity sets of the region they belong in.

* Adds support for checking the number of tiles a Mob occupies.

* Adds a ProjectileBuilder class to simplify creating projectiles.


Major-:
:tada:

PR #273, opened 2016-03-10
Simplify the method getIndefiniteArticle

CharlesVaneenoo:
Hello,
I found a redundant if in the method getIndefiniteArticle in the class LanguageUtil, so i remove it.
But something seems weird in the second if, the one i remove. Why is there just one '|' between first == 'l' and first == 'm' ?
Thanks.


CharlesVaneenoo:
@garyttierney bug fixed

PR #274, opened 2016-03-11
Adding special effects in food.rb

CharlesVaneenoo:
Hello,
I saw a TODO in food.rb so i added few special effects.
Thanks.


Major-:
Hi,

This won't actually work in the desired manner - this will just select e.g. a value between 5 and 7, and always use that randomly chosen value. The restoration stuff needs to be redone to take a lambda, which can be invoked every time the food is eaten.

Thanks anyway

PR #277, opened 2016-03-11
Complete dialogue in instructions.rb

CharlesVaneenoo:
Hello, 
I completed the TODO about skill_stats dialogue in instructions.rb.
What do you think about this ?
Thanks.

### DIFF
--- a/data/plugins/location/tutorial-island/instructions.rb
+++ b/data/plugins/location/tutorial-island/instructions.rb
@@ -126,7 +126,7 @@ module TutorialInstructions
       type :text
 
       title 'Your skill stats.'
-      text '' # TODO: this !!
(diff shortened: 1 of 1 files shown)


Major-:
The tutorial island plugin aims to be faithful to Runescape's original tutorial island, which includes using the exact text Runescape uses. I have a vague memory of this particular dialogue being more complicated than usual, but I can't remember why (and that might be completely wrong).

Feel free to open a new PR using the actual text, if you find it (I usually use [this video](https://www.youtube.com/watch?v=8ZoJuVbL1cU) when working on that plugin). If not then no problem, the tutorial island plugin is blocked on a couple of other things anyway

CharlesVaneenoo:
It's ok, I will work on it soon. 
Thanks.

PR #278, opened 2016-03-12
Add text in instruction.rb

CharlesVaneenoo:
Hello,
I added the text in instruction following the exact text of Runescape's original tutorial island.
Are you interested if I continue to add the following dialogue in the future ?
Thanks.


Major-:
Excellent, thanks. You're welcome to add the rest of the text if you like, but you might want to take a look at the rest of the plugin first, as getting all of the side-effects correct is quite complicated.

CharlesVaneenoo:
That's ok i will check the rest of the plugin first.
Thanks for the advice.

PR #284, opened 2016-03-15
Remove useless if

CharlesVaneenoo:
Hello, 
I saw two useless if in `GamePacketEncoder` in the package `org.apollo.net.codec.game`.
I decided to remove these two if to make the method *encode* more simpler.
I also pass a line (line 34) to make it more readable.
Thanks.


Major-:
Looks good, thanks

PR #285, opened 2016-03-15
improve putBits

CharlesVaneenoo:
Hello,
I thinked it will be better to put above the `tmp` variable. It clarify the function putBits and make it less redundant.
Thanks.


Major-:
Thanks!

PR #288, opened 2016-03-17
Turn player to objects when interacting with them

### LINKED ISSUE #271: Finding the center location of an object
doge

garyttierney:
Resolves #271 I believe. I don't think the issue there is that objects are larger than 1 tile, but that some plugins are handling turning to objects themselves.

### DIFF
--- a/data/plugins/entity/interaction/object.rb
+++ b/data/plugins/entity/interaction/object.rb
@@ -1,0 +1,6 @@
+# Catch all object actions and turn the player to
+# the position of the object
+
+on :message, :object_action do |player, message|
+  player.turn_to message.position
+end
--- a/data/plugins/entity/interaction/plugin.xml
+++ b/data/plugins/entity/interaction/plugin.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0"?>
+<plugin>
+    <id>entity-interaction</id>
+    <version>0.1</version>
+    <name>Entity Interaction</name>
+    <description>Adds support for generic entity interactions (facing objects when using an action, etc).</description>
+    <authors>
+        <author>garyttierney</author>
+    </authors>
+    <scripts>
+        <script>object.rb</script>
+    </scripts>
+</plugin>


ryleykimmel:
See my reply in #271 (also this wouldn't work for objects larger than 2 tiles, the 'center' is not always the actual position of the object)

Annnnd thinking about it objects have 'valid sides' where you can interact, this is the position you face as well, not the position from where you are currently standing. (i.e altar only from front)

WizardJesse1:
ryley is right about objects larger then 2 tiles

data/plugins/entity/interaction/object.rb
```diff
@@ -0,0 +1,6 @@
+# Catch all object actions and turn the player to
+# the position of the object
+
+on :message, :object_action do |player, message|
+  player.turn_to message.position
```
ryleykimmel: I think we need a method which determines the orientation we need to face and the position we need to face (the center of the object is not always the position it's spawned on). IIRC there is an opcode in the object definitions which holds orientation flags denoting what side of the object is intractable. I believe it's used in the clients route finding algorithm.

garyttierney: We currently have code for finding the origin and edges of an object, but having 'interactable sides' is news to me.  Got anywhere you can point to for this?

ryleykimmel: From Client#processClickedObject

```java
            int surroundings = definition.getSurroundings();
            if (orientation != 0) {
                surroundings = (surroundings << orientation & 0xF) + (surroundings >> 4 - orientation);
            }
```

The flag is passed to the walk method and checked in CollisionMap#reachedObejct

```java
    public boolean reachedObject(int x, int y, int finalX, int finalY, int height, int surroundings, int width) {
        int maxX = finalX + width - 1;
        int maxY = finalY + height - 1;

        if (x >= finalX && x <= maxX && y >= finalY && y <= maxY) {
            return true;
        } else if (x == finalX - 1 && y >= finalY && y <= maxY
                && (adjacencies[x - xOffset][y - yOffset] & WALL_EAST) == 0 && (surroundings & WALL_EAST) == 0) {
            return true;
        } else if (x == maxX + 1 && y >= finalY && y <= maxY && (adjacencies[x - xOffset][y - yOffset] & WALL_WEST) == 0
                && (surroundings & WALL_NORTH) == 0) {
            return true;
        } else if (y == finalY - 1 && x >= finalX && x <= maxX
                && (adjacencies[x - xOffset][y - yOffset] & WALL_NORTH) == 0 && (surroundings & WALL_NORTHEAST) == 0) {
            return true;
        }

        return y == maxY + 1 && x >= finalX && x <= maxX && (adjacencies[x - xOffset][y - yOffset] & WALL_SOUTH) == 0
                && (surroundings & WALL_NORTHWEST) == 0;
    }
```

garyttierney: Awesome!  Thanks for pointing that out.

@ryleykimmel++

Major-:
Is this still blocked on ryley's comment?

garyttierney:
@Major- no longer blocked, but needs reworked.  Server needs to use the surroundings information to determine where a client should walk to for following also.

PR #289, opened 2016-03-18
Rename get_next_dialogue

CharlesVaneenoo:
Hello,
I think this rename should be better for understanding.
Thanks.


Major-:
Maybe `decode_next_dialogue`? The fact that it's usually a Hash isn't too important (you usually wouldn't refer to types in a dynamic language like ruby).

CharlesVaneenoo:
I agree, I rename in `decode_next_dialogue`.

Major-:
Great, thanks!

PR #290, opened 2016-03-18
Remove useless TODO

KGALLET:
Hello,

I saw two TODO but for me there are already done. So i just remove the TO-DO annotation.

Thanks


Major-:
These comments are referring to the attribute system (with the current built-ins found [here](https://github.com/CharlesVaneenoo/apollo/blob/remote-useless-TODO/game/src/main/org/apollo/game/model/entity/Player.java#L76)), although they might be unnecessary anyway: they're both relevant to combat, so @garyttierney any thoughts?

PR #291, opened 2016-03-18
remove old TODO 

pheq:
Hi, we saw two TODO in instructions.rb related to the update of the island progress, they say that it has to be set on moving around but in game it actually does it. Is it an oversight or maybe I misunderstood something ?


Major-:
Nope you're absolutely right, I forgot to remove these. Thanks!

PR #292, opened 2016-03-18
Add an ore

CharlesVaneenoo:
Hello,
I added the ore pure essence and rune essence but i'm not sure about the `ESSENCE_OBJECTS`.
How do I know if it's correct or not.
Thanks.

### DIFF
--- a/data/plugins/skill/mining/ore.rb
+++ b/data/plugins/skill/mining/ore.rb
@@ -80,6 +80,10 @@ RUNITE_OBJECTS = {
   14_861 => 14_834
 }
 
+ESSENCE_OBJECTS = {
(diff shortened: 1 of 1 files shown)


Major-:
I think the essence stuff needs different behaviour, which is why the current TODO is there. I think @WizardJesse1 is doing a bunch of mining stuff now, any thoughts?

WizardJesse1:
I have essence done already and most of mining done I just need to finish testing

Major-:
Looks like Jesse has it covered then - thanks a lot anyway @CharlesVaneenoo, I really appreciate the effort you guys are going to, fixing and removing all the old TODOs.

PR #295, opened 2016-03-18
Fix Distance check

WizardJesse1:
Before:

http://recordit.co/qFCM2jaox2

After:

http://recordit.co/Jre4B5s1zT


Major-:
Yep looks good, nice catch.

Issue #247, opened 2016-02-25
Pickaxe Searching Mining


WizardJesse1:
doge

Major-:
When you close an issue please give a reason (e.g. "fixed in [commit]")

Major-:
@WizardJesse1 is this still an issue? i don't know what the original problem was

WizardJesse1:
@Major- it would be fixed in my mining if it was

tlf30:
@WizardJesse1 What was the original issue?
 I am currently working on the kotlin version of the plugin for mining and would like to know if there is a bug in the original ruby that might be pulled over into the kotlin version.

Major-:
Issue was that you wouldn't use the best pickaxe you had, it would use whatever it found first

tlf30:
Oh OK. That is fixed in the kotlin version I wrote.

tlf30:
I think this can be closed now as it is fixed and merged into the new kotlin plugin system.

Major-:
@tlf30 Excellent, thanks!

Issue #184, opened 2016-02-09
Add support for mining Rune and Pure essence


Major-:
This might be slightly tricky to do with the existing code, so d-medium.

tlf30:
I will look into this.

Issue #183, opened 2016-02-09
Don't guarantee a player will always successfully mine ore


Major-:
Currently the [mining plugin](https://github.com/apollo-rsps/apollo/blob/master/data/plugins/skill/mining/mining.rb#L65) always guarantees a player will successfully mine a rock, whereas in RS this usually takes several ticks at minimum (and typically much longer for higher level rocks such as rune). We need some sort of probability function.

tlf30:
Fixed in Kotlin port of plugin

tlf30:
I think this can be closed now as it is fixed and merged into the new kotlin plugin system.

Major-:
@tlf30 Great, thanks a lot!

Issue #182, opened 2016-02-09
Replace mined rocks with their empty equivalents


Major-:
Currently the mining plugin does not replace rocks that have had ore mined from them, meaning you can endlessly mine from the same rock with no time limits.

This will also require some sort of Task to replace expired rocks with their ore-containing equivalents, after the appropriate period of time has elapsed.

tlf30:
I think this can be closed now as it is fixed and merged into the new kotlin plugin system.

Major-:
Fixed in e61f4ef.

PR #297, opened 2016-03-20
Mining

WizardJesse1:
* Full object restoring and depleting (Fix Issue #182)

* Pickaxes now go to the highest one (Fix Issue #247)

* Full essence mining support (Fix Issue #184)

* Object sizes for the rocks are checked from the definition now

* Still needs work but players have a chance to fail mining ores (Fix Issue #183)

* Sand stone & Granite support with randomize rocks

* Expired mining check added

* Stopping the action is not based from when you get the ore but now from when the ore depletes

* Supports all message types from regular rocks to ores. 

* Pickaxes have a rate added to them since its based off the counter pulse anymore.


data/plugins/skill/mining/mining.rb
```diff
@@ -39,11 +50,29 @@ def start_mine(pickaxe)
   end
 
   def executeAction
-    skills = mob.skill_set
-    level = skills.get_skill(Skill::MINING).current_level
+    level = mob.skill_set.get_skill(Skill::MINING).current_level
     pickaxe = find_pickaxe
+
+    ores = @ore.id.reject { |ore| ore.level > level}
+    ore_id = if object_id == 2491
```
garyttierney: This isn't great, what object is the object with an id of 2491? Also, this would look cleaner if it were a simple conditional without the modifier.

data/plugins/skill/mining/pickaxe.rb
```diff
@@ -7,27 +7,28 @@
...
 end
 
 # NOTE: ADD LOWER LEVEL PICKAXES FIRST
-append_pickaxe(Pickaxe.new(1265, 1,  625, 8)) # bronze pickaxe
-append_pickaxe(Pickaxe.new(1267, 1,  626, 7)) # iron pickaxe
-append_pickaxe(Pickaxe.new(1269, 1,  627, 6)) # steel pickaxe
-append_pickaxe(Pickaxe.new(1273, 21, 629, 5)) # mithril pickaxe
-append_pickaxe(Pickaxe.new(1271, 31, 628, 4)) # adamant pickaxe
-append_pickaxe(Pickaxe.new(1275, 41, 624, 3)) # rune pickaxe
+append_pickaxe(Pickaxe.new(1265, 1,  625, 8, 0.05)) # bronze pickaxe
```
garyttierney: Could you change this to take named parameters like you have done with `gem.rb`?

data/plugins/skill/mining/mining.rb
WizardJesse1: pure essence is object 2491

PR #299, opened 2016-03-23
Add message when encoder is equals to null

CharlesVaneenoo:
Hello,
I saw that decode method print a message when the decoder is null in `GameMessageDecoder`.
So I added a message when our variable encoder is null.
Thanks.


garyttierney:
Hi,

Good change. Could save  wasted time debugging why some message isn't appearing on the client end. One nit pick though, you should use a `Logger` instead of printing to stdout. Besides that looks good.

Cheers.

garyttierney:
Are there going to be any more changes to this branch?

CharlesVaneenoo:
Not for the moment, maybe i will try to use Logger soon but i don't know when. Should I close this PR or you'll merge it like that ?

garyttierney:
Closing for now.

PR #260, opened 2016-02-29
Add support for combat

garyttierney:
:warning: not ready for pulling or code review, still cherry picking :warning:

PR #302, opened 2016-06-01
Mob following

stevesoltys:
Add support for mob following. Currently a WIP.
- Support for Mobs that have a size greater than one tile is still untested, but _should_ work. See `get_minimum_distance` in `FollowingModule`.
- Pathfinding seemingly needs some work, as it's easy to get a `Mob` stuck behind a wall, even while using the `AStarPathfindingAlgorithm` implementation.
- Support for projectiles is also not there, but will be a simple addition once projectile clipping has been completed. See `follow` in `FollowingModule`.

### DIFF
--- a/data/plugins/follow/util.rb
+++ b/data/plugins/follow/util.rb
@@ -1,5 +1,5 @@
 java_import 'org.apollo.game.model.entity.path.AStarPathfindingAlgorithm'
-java_import 'org.apollo.game.model.entity.path.ChebyshevHeuristic'
+java_import 'org.apollo.game.model.entity.path.EuclideanHeuristic'
 java_import 'org.apollo.game.model.entity.path.SimplePathfindingAlgorithm'
 java_import 'org.apollo.game.model.entity.Npc'
 java_import 'org.apollo.game.model.Direction'
@@ -16,7 +16,7 @@ module FollowingModule
 
   ##
   # A pathfinder used for precise following (i.e.: A player attacking another player).
-  FOLLOW_PATH_FINDER = AStarPathfindingAlgorithm.new($world, ChebyshevHeuristic.new)
+  FOLLOW_PATH_FINDER = AStarPathfindingAlgorithm.new($world, EuclideanHeuristic.new)
 
   ##
   # Gets the minimum distance that a mob must be at while following another.
@@ -110,7 +110,7 @@ module FollowingModule
         reset(mob) && task.stop && next
       end
 
-      next unless mob.walking_queue.size == 0
+      next unless mob.walking_queue.size <= 1
 
       # the minimum distance that we must keep between us and the mob we're following
       minimum_distance = get_minimum_distance(mob, other_mob)
@@ -123,7 +123,7 @@ module FollowingModule
       end_position = nil
 
       if distance < minimum_distance
-        end_position = other_mob.position.from_direction(HORIZONTAL_DIRECTIONS.sample)
+        end_position = other_mob.position.add(HORIZONTAL_DIRECTIONS.sample)
 
       elsif distance > max_distance
         end_position = other_mob.position
--- a/data/plugins/follow/util.rb
+++ b/data/plugins/follow/util.rb
@@ -1,0 +1,139 @@
+java_import 'org.apollo.game.model.entity.path.AStarPathfindingAlgorithm'
+java_import 'org.apollo.game.model.entity.path.ChebyshevHeuristic'
+java_import 'org.apollo.game.model.entity.path.SimplePathfindingAlgorithm'
+java_import 'org.apollo.game.model.entity.Npc'
+java_import 'org.apollo.game.model.Direction'
+
+module FollowingModule
+
+  ##
+  # An array containing the north, east, south, and west directions.
+  HORIZONTAL_DIRECTIONS = [Direction::NORTH, Direction::EAST, Direction::SOUTH, Direction::WEST]
+
+  ##
+  # A pathfinder used for simple following (i.e.: An NPC attacking a player).
+  SIMPLE_PATH_FINDER = SimplePathfindingAlgorithm.new($world)
+
+  ##
+  # A pathfinder used for precise following (i.e.: A player attacking another player).
+  FOLLOW_PATH_FINDER = AStarPathfindingAlgorithm.new($world, ChebyshevHeuristic.new)
+
+  ##
+  # Gets the minimum distance that a mob must be at while following another.
+  # TODO: This function should work well for following larger entities, but it needs to be cleaned up.
+  #
+  # @param [Mob] mob The following mob.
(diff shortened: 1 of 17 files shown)


game/src/main/org/apollo/game/model/Position.java
```diff
@@ -224,6 +224,66 @@ public boolean isWithinDistance(Position other, int distance) {
...
 	}
 
+	/**
+	 * Gets a position, based upon this position, from the given deltas.
+	 *
+	 * @param deltaX The delta x.
+	 * @param deltaY The delta y.
+	 * @return The position.
+	 */
+	public Position fromDeltas(int deltaX, int deltaY) {
```
garyttierney:
```
public Position add(int x, int y) {
```

game/src/main/org/apollo/game/model/Position.java
```diff
@@ -224,6 +224,66 @@ public boolean isWithinDistance(Position other, int distance) {
...
+		return new Position(getX() + deltaX, getY() + deltaY, getHeight() + deltaZ);
+	}
+
+	/**
+	 * Gets the adjacent position in the given direction.
+	 *
+	 * @param direction the direction
+	 * @return the position
+	 */
+	public Position fromDirection(Direction direction) {
```
garyttierney:
```
public Position add(Direction direction) {
```
Also use the helper functions in Direction: https://github.com/apollo-rsps/apollo/blob/bugfix/collision-detection/game/src/main/org/apollo/game/model/Direction.java#L190

data/plugins/follow/util.rb
```diff
@@ -0,0 +1,139 @@
...
+  def self.follow(mob, other_mob, maximum_distance, simple, projectile_clipped)
+
+    mob.following = other_mob.index
+    mob.interacting_mob = other_mob
+
+    schedule 0, true do |task|
+
+      # if we're following a different entity, stop this task
+      # TODO: Could the indices of NPCs and Players collide since they are in different instances of MobRepository..?
+      if mob.following != other_mob.index
```
garyttierney: This is a valid point. You might want to store a reference on the Mob rather than an index as an attribtue.

WizardJesse1:
http://recordit.co/Lf1H9RJktV any idea why its always trying to go to the side of the player ;o?

garyttierney:
@WizardJesse1 I believe this is in reference to that:

```
21:09 < shiver> it's not identical to RS atm tho
21:09 < shiver> like, it doesn't do the whole dance thing
21:09 < sfix> o
21:09 < sfix> is it still a WIP?
21:09 < shiver> i have the function for larger entities too
21:09 < shiver> but i haven't tested it
21:09 < sfix> that's a prereq for merging
21:09 < shiver> yea, but it works. maybe merge it but keep the issue?
21:10 < sfix> would rather wait until the feature is complete
21:10 < sfix> but anyway make the PR
21:10 < sfix> can get a few more set of eyes on it
```

garyttierney:
Looks good so far. Only real missing piece of the puzzle is finding out what direction a Mob is currently facing.

stevesoltys:
@WizardJesse1 I think that's an issue with the A* pathfinding implementation, rather than this plugin.

WizardJesse1:
@stevesoltys alright then, other then that it seems pretty good

data/plugins/follow/util.rb
```diff
@@ -0,0 +1,139 @@
...
+  # An array containing the north, east, south, and west directions.
+  HORIZONTAL_DIRECTIONS = [Direction::NORTH, Direction::EAST, Direction::SOUTH, Direction::WEST]
+
+  ##
+  # A pathfinder used for simple following (i.e.: An NPC attacking a player).
+  SIMPLE_PATH_FINDER = SimplePathfindingAlgorithm.new($world)
+
+  ##
+  # A pathfinder used for precise following (i.e.: A player attacking another player).
+  FOLLOW_PATH_FINDER = AStarPathfindingAlgorithm.new($world, ChebyshevHeuristic.new)
```
garyttierney: Use the `EuclideanHeuristic` implementation (or fix Chebyshev)

data/plugins/follow/util.rb
stevesoltys: Currently, attributes are limited to certain data types. See `AttributeType`.

Any other suggestions? @garyttierney

stevesoltys:
Made most changes @garyttierney, only the collision issue remains.

Following single tile mobs seems perfect, besides the inability to "dance". Please confirm that it works just as well on your end @WizardJesse1

WizardJesse1:
@stevesoltys it still seems to be holding diagonal http://recordit.co/TqilAsR40X it doesn't happen all the time but it does happen

data/plugins/follow/util.rb
garyttierney: @stevesoltys something similar to combat maybe https://github.com/apollo-rsps/apollo/pull/260/files#diff-70a326256cf9b797a8a087b495c01987R3

garyttierney:
@WizardJesse1 this might be down to not trying to get behind the player. That's still one thing to do IIRC.

WizardJesse1:
o kk then @garyttierney

garyttierney:
Latest changes look good. However, would still like the 'get behind' / dancing behavior in. Is this something you can handle @stevesoltys ?

stevesoltys:
@garyttierney, going to work on getting that behavior in right now.  Also, mobs that are larger than one tile still need to be tested (and looking back at it now, they probably do not currently work).

stevesoltys:
[https://cdn.streamable.com/video/mp4/0npo.mp4](url)

Will be committing these changes soon.

Lmctruck30:
Issue with this is doors now. Doors are now blocking the walking path when opened. Also in your Ruby you put to listen for a player action. As it stands it is listening for all player actions not just the follow. So if you trade someone you will follow.

garyttierney:
World#intersects() is a hack and incorrect, it doesn't correctly account for
paths that don't have equal X and Y deltas.

The best thing to do here is register listeners with a Region which can update
the CollisionMatrix when a door is opened / closed (though, one listener
should be able to do this for any type of dynamic object).

I'm going to look at some of this over the festive period.

Lmctruck30:
@garyttierney Thank you. Be awesome to see this system finished =D

Lmctruck30:
@garyttierney Is there an update on this PR?

garyttierney:
@Lmctruck30 I've done some more work on this to accommodate for dynamic objects.  I need to wire up the new collision matrix code to some region listeners for adding / removing objects and it should be all set.

Lmctruck30:
@garyttierney let me know if you want me to test.

garyttierney:
Closed by https://github.com/apollo-rsps/apollo/commit/e5a6638e2fc0bc4f524906639c42893998c4e594.

Issue #135, opened 2016-01-26
Replace maven with Gradle


Major-:
Gradle offers much greater potential for customization, and better multi-language integration (which is going to be very useful when we actually have unit tests for ruby). 

Shouldn't be too tricky to port from maven to gradle.

Major-:
Interested in doing this @garyttierney?

Major-:
We want:

- [ ] Java test support (which is supported natively by gradle)
- [ ] Ruby test support (need to decide how to do this)
- [ ] Rubocop support

garyttierney:
Closed in https://github.com/apollo-rsps/apollo/commit/d8f4aa88fe82c1725063a350d8be49e7abc0c12e.

PR #309, opened 2017-01-02
Port the Maven POMs to Gradle build scripts

garyttierney:
This updates Apollo to use Gradle as a build system instead of Maven.  This is significantly easier to configure and friendlier for end-users, too.  To run the server now, instead of running `mvn install` and `mvn exec:java` you should do:

```shell
$ gradle run
```

Which will compile and boot Apollo with the default command line and VM arguments (no command line, VM -Xmx750M)


garyttierney:
Note that the Travis-CI configuration doesn't need updated since it automatically detects Gradle build scripts and runs 'gradle check'.  Merging this to close #135.

PR #311, opened 2017-02-04
Add unit tests for MessageHandlerChain and MessageHandlerChainSet

BeeHiveJava:
Hey there,

I feel like the two mentioned classes are complex enough to have some tests added to them. I tested the following behaviour:

In ```MessageHandlerChainSet``` and ```MessageHandlerChain```:
- ```notify(..)``` should invoke the ```handle(..)``` method in the correct ```MessageHandler```.
- ```notify(..)``` should return ```true``` if a handled ```Message``` does not get terminted.
- ```notify(..)``` should not invoke a ```MessageHandler``` if a ```Message``` was terminated before ```handle(..)``` was called. (Note that I think no ```MessageHandler``` should be called at all if a ```Message``` gets terminated before it was even handled. This happens due to the if-statement being placed after the ```handle(..)``` call ([link to the if-statement](https://github.com/apollo-rsps/apollo/blob/master/game/src/main/org/apollo/game/message/handler/MessageHandlerChain.java#L59))
- ```notify(..)``` should return ```false``` if a handled ```Message``` gets terminated.

Also, I chose to manually 'mock' a ```Message``` and ```MessageHandler``` in order to create better readability. If you'd like me to use PowerMock for that, do tell.

So please have a look at whether a ```Message``` should get handled at all if it was terminated in the first place.

Thanks,

Lesley


garyttierney:
@BeeHiveJava Apologies for not seeing this sooner!  I'll have a look at it now.

PR #320, opened 2017-05-24
Broken Method for getting magic level.

Promises:


Promises:
Should be maximum level not current, sorry.

Major-:
@Promises I think current is right? If the player's level is debuffed then they can't use a spell, I think?

From my understanding the `current_level` change is good.

Promises:
Ye sure, now that I'm not on my 10 th cup of Joe, you are right.

Issue #328, opened 2017-05-25
Investigate sharp increase in memory usage


Major-:
At some point in the not-so-distant past Apollo would function fine with `-Xmx750M`, but now needs `-Xmx1750M` to even start. Is this just a consequence of moving from 317 to 377 (and hence a large increase in the object data we load and store in-memory)?

kylestev:
Not sure how much this helps but here's a 10s snapshot of the memory usage during boot

![image](https://user-images.githubusercontent.com/819571/30549031-84d03778-9c48-11e7-9767-bf903462b747.png)

garyttierney:
What you're probably seeing there is the collision maps being built, and the interim objects created to do that being GC'd.  @Major- I think we definitely need to lazy load  these.  A basic implementation could be as simple as preventing movement 'til a collision map is loaded.

kylestev:
Perhaps some profiling (not something I have experience with) is in order to make sure the right parts are optimized.

rmcmk:
I vote for development environment only lazy loading for pretty much everything to keep startup time low. Very necessary when developing, especially if we cannot feasibly come up with a runtime plugin reloading strategy. A simle caching mechanism would suffice, I hear guava has a nice LRU cache implementation

dosier:
I managed to reduce the startup time with roughly 3 seconds by wrapping two of `Preconditions` checks inside an if statement both in `CollisionMatrix.java`
```java
	private int indexOf(int x, int y) {
		if (x < 0 || x >= width)
			Preconditions.checkElementIndex(x, width, "X coordinate must be [0, " + width + "), received " + x + ".");
		if (y < 0 || y >= width)
			Preconditions.checkElementIndex(y, length, "Y coordinate must be [0, " + length + "), received " + y + ".");
		return y * width + x;
	}
```
and in `Region.java`
```java
	public CollisionMatrix getMatrix(int height) {
		if (height < 0 || height >= matrices.length)
			Preconditions.checkElementIndex(height, matrices.length, String.format(ILLEGAL_MATRIX_HEIGHT, matrices.length, height));
		return matrices[height];
	}
```
This prevents redundant string creations.

PR #329, opened 2017-05-25
Increase maximum heap size

Major-:
Required or Apollo reaches the limit on boot. See #328.


garyttierney:
NACK.  This is a gradle bug.  I have a workaround for this.

PR #336, opened 2017-05-26
Make immutable DistancedAction fields visible to subclasses

Major-:


game/src/main/org/apollo/game/action/DistancedAction.java
```diff
@@ -15,27 +15,27 @@
...
 	 * The position to distance check with.
 	 */
-	private final Position position;
+	protected final Position position;
 
 	/**
 	 * A flag indicating if the distance has been reached yet.
 	 */
-	private boolean reached = false;
+	private boolean reached;
```
Major-: Ya but I'm unwilling to make a separate commit for a single-line stylistic change

PR #338, opened 2017-05-30
Replace Ruby with Kotlin

garyttierney:
As discussed in #310 this pull request introduces a new Kotlin based plugin system as well as a much, much better plugin build system.

Major planned features:

- [x] Debugging support (with IntelliJ or Eclipse's step-through debugger for example)
- [x] Unit test support
- [x] Integration test framework
- [x] Full editor integration (auto-completion, error reporting, etc.)

For a more detailed explanation of what we get with kotlin and this new plugin system see the discussion in #310.

As well as the plugin system itself, there is a lot of plugins written in Ruby left to port.  A list of these is below.

- [x] areas
- [x] bank
- chat
  - [x] privacy
  - [x] private-messaging
- cmd
  - [x] animate
  - [x] bank
  - [x] item
  - [x] lookup
  - [x] messaging
  - [x] npc
  - [x] punishment
  - [x] skill
  - [x] teleport
- [x] consumables
- [x] dummy
- [x] emote-tab
- entity
  - [x] attributes
  - mob
    - [x] extension
    - [x] following
    - [x] walk-to
  - [x] spawning
- location
  - [x] al-kharid
  - [x] edgeville
  - [x] falador
  - [x] lumbridge
  - [x] varrock
- [x] logout
- navigation
  - [x] door
- [x] player-action
- [x] quest
- [x] run
- [x] shops
- skill
  - [x] fishing
  - [x] herblore
  - [x] mining
  - [x] prayer
  - [x] runecraft
- [x] util (now part of the api plugin)

Two pugins have been deselected from the kotlin port, and will be done at a later date - dialogue and the (non-combat) magic skill.

This doesn't include tests for the plugins - see #361.

PR #346, opened 2017-08-18
Added mining plugin for Kotlin

tlf30:
Hello,
I have ported the mining plugin to Kotlin. This is my very first time ever working in Kotlin, so if you have any feed back I am all ears. 

I do not have a working test environment setup for the Kotlin plugin version of Apollo, I do believe that this code is correct, but I would greatly appreciate if someone could test this for me. 

If this all looks good, I will work on porting over other plugins. 

I hope this is usable and helps!
~Trevor


tlf30:
I do not know why the gem.rb and respawn.rb were originally in the mining plugin dir, but I have also included gem in my port as gem.kt. I did not port respawn because I am not sure why it is there or what it is used for.

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -0,0 +1,129 @@
...
+        .where {option == 2}
+        .then {
+            if (ORES.contains(id)) {
+
+            } else if (EXPIRED_ORES.contains(id)) {
+                it.startAction(ExpiredProspectingAction(it, this.position))
+            }
+        }
+
+//Init the ore dict when we start the server. Probably better ways to do this, but I do not know them... Sorry.
```
garyttierney: No, this is fine.  Having start-up methods is an intentional design decision.

garyttierney:
Thanks for the contribution.  Will get around to reviewing it tonight.

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: OK, thanks for letting me know! I'll change the comment.

tlf30:
Is it OK if I add fishing to this pull request?

garyttierney:
Please make a separate pull request for fishing.

tlf30:
OK, I will do.

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -0,0 +1,130 @@
...
+
+        //check that our pick can mine the ore
+        if (pickaxe == null || level < pickaxe.level) {
+            mob.sendMessage("You do not have a pickaxe for which you have the level to use.")
+            stop()
+            return
+        }
+
+        //check that we can mine the ore
+        if (ore.level > level) {
```
garyttierney: Be consistent with the `pickaxe.level` test.  `level < ore.level`.

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -0,0 +1,130 @@
...
+            mob.sendMessage("You do not have the required level to mine this rock.")
+            stop()
+            return
+        }
+
+        //start the process of mining
+        if (started) {
+            if (counter == 0) {
+                //TODO: calculate the chance that the player can actually get the ore
```
garyttierney: I think this needs to be done before we can merge this.

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -0,0 +1,130 @@
...
+                }
+            }
+            counter -= 1
+        } else {
+            startMine(pickaxe)
+        }
+    }
+
+    fun findPickaxe(): Pickaxe? {
+        var holding = mob.equipment.get(EquipmentConstants.WEAPON)
```
garyttierney: I think this method needs to be a little smarter.  When you mine an ore I'm sure that you use the best pickaxe available to you.

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -0,0 +1,130 @@
...
+        counter = pickaxe.pulses
+    }
+}
+
+class ExpiredProspectingAction : DistancedAction<Player> {
+
+    constructor(mob: Player, position: Position) : super(0, true, mob, position, 1 /* ORE SIZE */)
+
+    override fun executeAction() {
+        mob.sendMessage("There is currently no ore avalible in this rock.")
```
garyttierney: Typo here.

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -0,0 +1,130 @@
...
+    var started = false;
+    var ore: Ore by Delegates.notNull()
+
+    constructor(mob: Player, position: Position, ore: Ore) : super(3 /* PROSPECT PULSES */, true, mob, position, 1 /* ORE SIZE */) {
+        this.ore = ore;
+    }
+
+    override fun executeAction() {
+        if (started) {
+            mob.sendMessage("This rock contains " + ItemDefinition.lookup(ore.id).name.substring(3).toLowerCase() + ".")
```
garyttierney: As in fishing review, use `lookup` method from the utils plugin.

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: Does anyone know the chance that the player gets the ore? It was not in the ruby version of this plugin.

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: This function actually does that already. The picks that it iterates over are in order of best to worst.

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: Thanks!

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: Thanks!

tlf30:
Still need to get lookup working and get a chance for ore being mined.

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: Maybe I am not understanding the lookup util, but it appears to use the name to get the Item obj

garyttierney: Sorry, my mistake.  That's missing functionality on my part.  I'll correct this later.

game/src/plugins/skills/mining/src/mining.plugin.kts
garyttierney: Some info on that can be found here: http://runescape.wikia.com/wiki/Talk:Mining#Mining_success_rate_formula

game/src/plugins/skills/mining/src/mining.plugin.kts
garyttierney: Yep, but it doesn't seem to take the players level into account.  i.e., if they have a rune pickaxe in their inventory but can only use adamant then findPickaxe() should return the adamant pickaxe.

tlf30: Oh, I forgot about that. Thank you!!

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: OK, sounds good

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: OK, implemented! Thank you, I was looking all over for this!

tlf30:
OK, I think I have everything working on this now...?
I have implemented the algorithm for determining the chance to mine an ore. 
i also fixed not checking if the player can use the pickaxes it has.

The only thing left is to figure out how to expire the ore. I would really appreciate if someone could point me in the right direction on this one.  

Thank you,
Trevor

garyttierney:
>The only thing left is to figure out how to expire the ore. I would really appreciate if someone could point me in the right direction on this one.

cc @Major- 

is the best way of doing this still scheduling a task that immediately removes the object, replaces it, and does the same again after the respawn time?

tlf30:
Also, are we respawning ores based on server pop or constant times? I know it used to be pop based but then was changed to be constant. 
http://www.runescape-static.salmoneus.net/skills/mining.html#ore_respawn_times

garyttierney:
@tlf30 lets go with constant time for now.  If anyone running apollo sees this to be a problem we can change it.

tlf30:
OK, I have tasks and the respawn times figured out, the last question I have for now is how do I remove/replace an object?

tlf30:
So I think I have found a way to remove and respawn the entities for the rocks. I would love some feedback on the approach though.

tlf30:
OK, I have a problem in my implementation. 
1. If two entities are at the same location, the ore is not respawned and not replaced by the depleted ore. It instead disappears permanently.   
2. Not all ores are at the location stored in position for the event. Which is all the more questionable. 

To test these, go to 3083 9800, give yourself 1265, and mine some of the tin ore.  

Other than these two problems, the code works. I am looking into solutions, but do not have any at this time. Also this code contains some debugging print statements.

Lastly, and unrelated. On my client I am not getting messages for the player from the server. It is very likely an issue with my client, but I do not know for sure.

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -37,7 +37,7 @@ class MiningAction(val player: Player, val p: Position, val ore: Ore) : Distance
             if (counter == 0) {
                 //TODO: calculate the chance that the player can actually get the ore
                 if (mob.inventory.add(ore.id)) {
-                    mob.sendMessage("You managed to mine some " + ItemDefinition.lookup(ore.id).name.substring(3).toLowerCase() + ".")
+                    mob.sendMessage("You managed to mine some $(ItemDefinition.lookup(ore.id).name.substring(3).toLowerCase()).")
```
Major-: Can you put the name into a variable above? It's a bit too long to interpolate

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -79,17 +74,11 @@ class ExpiredProspectingAction : DistancedAction<Player> {
     override fun executeAction() {
         mob.sendMessage("There is currently no ore available in this rock.")
     }
-
 }
 
-class ProspectingAction : DistancedAction<Player> {
+class ProspectingAction(val m: Player, val p: Position, val ore: Ore) : DistancedAction<Player>(3 /* PROSPECT PULSES */, true, m, p, 1 /* ORE SIZE */) {
```
Major-: Can you put the constants (pulses/ore size) into named values? To get constants in Kotlin you use [companion objects:](https://kotlinlang.org/docs/reference/object-declarations.html#companion-objects):

```kotlin
class ProspectingAction ..........

    companion object {
        private val PROSPECT_PULSES = 3
    }
```

game/src/plugins/skills/mining/src/gem.kt
```diff
@@ -2,9 +2,12 @@ package org.apollo.game.plugin.skills.mining
 
 data class Gem(val id: Int, val chance: Int)
```
Major-: We should use enums for gems/pickaxes

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -50,23 +60,33 @@ class MiningAction(val player: Player, val p: Position, val ore: Ore) : Distance
         }
     }
 
-    fun findPickaxe(): Pickaxe? { // Find the best pick the player has
+    private fun findPickaxe(): Pickaxe? {
         for (id in getPickaxes()) {
+            val pick = lookupPickaxe(id) //Will not return null
```
Major-: How do we know this won't return `null`?

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -50,23 +60,33 @@ class MiningAction(val player: Player, val p: Position, val ore: Ore) : Distance
...
+            if (pick!!.level > mob.skillSet.getSkill(Skill.MINING).currentLevel) {
```
Major-: Statically importing `MINING` and `WEAPON` here would be nice

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -50,23 +60,33 @@ class MiningAction(val player: Player, val p: Position, val ore: Ore) : Distance
...
+    private fun startMine(pickaxe: Pickaxe) {
         started = true
         mob.sendMessage("You swing your pick at the rock.")
         mob.playAnimation(pickaxe.animation)
         counter = pickaxe.pulses
     }
+
+    private fun miningSuccessful(oreChance: Double, oreChanceOffset: Boolean, playerLevel: Int): Boolean {
+        val percent = (oreChance * playerLevel + (if (oreChanceOffset) 1 else 0 )) * 100;
+        //See: http://runescape.wikia.com/wiki/Talk:Mining#Mining_success_rate_formula
```
Major-: Can you add this link as a doc comment to the function?

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -50,23 +60,33 @@ class MiningAction(val player: Player, val p: Position, val ore: Ore) : Distance
...
-    fun startMine(pickaxe: Pickaxe) {
+    private fun startMine(pickaxe: Pickaxe) {
         started = true
         mob.sendMessage("You swing your pick at the rock.")
         mob.playAnimation(pickaxe.animation)
         counter = pickaxe.pulses
     }
+
+    private fun miningSuccessful(oreChance: Double, oreChanceOffset: Boolean, playerLevel: Int): Boolean {
+        val percent = (oreChance * playerLevel + (if (oreChanceOffset) 1 else 0 )) * 100;
```
Major-: Can you move the if/else to a separate value please

game/src/plugins/skills/mining/src/ore.kt
```diff
@@ -9,7 +9,7 @@ Thanks to Clifton <http://www.rune-server.org/members/clifton/> for helping
 to find some of the expired object IDs.
  */
 
-data class Ore(val id: Int, val objects: Map<Int, Int>, val level: Int, val exp: Double, val respawn: Int)
+data class Ore(val id: Int, val objects: Map<Int, Int>, val level: Int, val exp: Double, val respawn: Int, val chance: Double, val chanceOffset: Boolean)
```
Major-: Enum please

Major-:
@tlf30 can you clarify what the problem (or problems) are? Also some good work, thanks!

tlf30:
@Major- Thank you for the feed back! 
The issues that I am having:
1. Some ore rocks disappear completely after being mined 
2. Some ore rocks disappear instead of becoming the depleted rock, but will show when respawned as a normal rock.     
3. Some ore rocks correctly become depleted but disappear instead of being respawned as a normal rock.

The ore rocks that have these problems are always the same physical ones in the world. It is not related to the ID of the ore rock as far as I can tell. The same rock always has the same issue. I have been doing my testing in the mine at 3083 9800. I also tested two different clients which both have the same issue with the same rocks.

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: Because getPickaxes returns the set of ID values that lookupPickaxe looks too.

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: Will do

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: OK

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: OK

game/src/plugins/skills/mining/src/mining.plugin.kts
tlf30: Will do

game/src/plugins/skills/mining/src/gem.kt
tlf30: Pickaxes and Gems can be enums. But I feel it over complicates things a little.

game/src/plugins/skills/mining/src/ore.kt
tlf30: Again I feel like this over complicates things.

garyttierney: Think about it from a development experience perspective: now any time someone wants to use an `Ore` their IDE will make all these values available to them via autocomplete.

tlf30: I agree it looks better.

Major-:
I guess that's a problem with the region/object management system then

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -126,7 +126,8 @@ class MiningAction(val player: Player, val objectID: Int, val p: Position, val o
             if (pick!!.level > mob.skillSet.getSkill(Skill.MINING).currentLevel) {
                 continue;
             }
-            if (mob.equipment.get(EquipmentConstants.WEAPON).id == pick.id) {
+            val weponSlot = mob.equipment.get(EquipmentConstants.WEAPON)
```
Major-: `weapon`

game/src/plugins/skills/mining/src/mining.plugin.kts
```diff
@@ -198,8 +198,8 @@ class ProspectingAction(val m: Player, val p: Position, val ore: Ore) : Distance
 on {ObjectActionMessage::class}
         .where {option == 1}
         .then {
-            if (ORES.contains(id)) {
-                it.startAction(MiningAction(it, id, this.position, ORES.get(id)!!))
+            if (lookupOreRock(id) != null) {
+                it.startAction(MiningAction(it, id, this.position, lookupOreRock(id)!!))
             } else {
                 System.out.println("Unknown ore: " + id)
```
Major-: Can we remove this debug message? If the ore doesn't exist then it means the player clicked a different object, we don't want to spam the console every time that happens

game/src/plugins/skills/mining/src/gem.kt
```diff
@@ -1,13 +1,18 @@
 package org.apollo.game.plugin.skills.mining
 
-data class Gem(val id: Int, val chance: Int)
-
-private val GEMS = mutableMapOf<Int, Gem>()
+enum class Gem(val id: Int, val chance: Int) {
+    UNCUT_SAPPHIRE(1623, 0), // uncut sapphire
```
Major-: You can delete these comments - if you're going to add comment, remember to ask yourself "does this add value/meaning to the code?"

game/src/plugins/skills/mining/src/gem.kt
```diff
@@ -1,13 +1,18 @@
...
+    UNCUT_EMERALD(1605, 0),// uncut emerald
+    UNCUT_RUBY(1619, 0), // uncut ruby
+    UNCUT_DIAMOND(1617, 0)  // uncut diamond
+}
 
 fun lookupGem(id: Int): Gem? {
-    return GEMS[id]
+    for (gem in Gem.values()) {
```
Major-: Try and use kotlin features when writing in kotlin, like replacing this entire function with:

```kotlin
fun lookupGem(id: Int): Gem? = Gem.values().find { it.id == id }
```

by combining a one-line function definition with kotlin's functional stuff

tlf30:
Thank you for the review @Major-, I was going to remove the debugging statements and clean up the comments after I had finished debugging why objects were disappearing. But if it is not an issue with this code I will clean it up now.

Thank you for all the help!

game/src/plugins/skills/mining/src/gem.kt
tlf30: Thanks. I did not know about this. I am still trying to get used to all the one liners in Kotlin.

Major-:
Can you give me some exact steps to see the object replacing bug (like coordinates to stand at and which rock to mine)? Does it always happen?

tlf30:
Teleport to 3038 9800. Give your self a pickaxe. Mine nearest coal coal object. It will become a depleted ore correctly but vanish when it should have become a coal rock again. Then if you go north east, there are 2 stone rocks. Mine both of them. One will vanish permanently. Other ore rocks will do other similar things if you mine them, but it is always the same rocks, not by id, but by location. 

commands:
```
::tele 3038 9800
::item 1275
::level 14 99 
```

game/src/plugins/skills/mining/src/gem.kt
Joshua-F: @Major- Wouldn't Gem.values() create a new array every time `lookupGem` is called then?

garyttierney: Yes.  Arrays of enum values should be cached.

tlf30: So does a global array need to be created to store the array from the enum? What would be the best way to cache it? If we still have to populate an array when using an enum, what advantage is gained by using them over a data class? We will now have two copies of the data instead of one.

I would also like to note that an enum generates a class at compile time so both and enum and data class are stored as objects and will take the same amount of space in memory when storing the same data.

Joshua-F: I don't personally know Kotlin, but in Java you typically do `public static final Gem VALUES = Gem.values()`.

tlf30: So for the sake of getting this done with I just created a val and stored the array.

garyttierney:
Merged in https://github.com/apollo-rsps/apollo/pull/338/commits/a54485e2637127070507802fdf6a5b45c8646e61.  Thanks!

game/src/plugins/skills/mining/src/gem.kt
Promises: amulet of glory with charges should effect chances of mining gems.

tlf30: There are several things that will. But at the moment we are just working on base functionality.

PR #350, opened 2017-09-11
Added woodcutting plugin for Kotlin

tlf30:
This is a wood cutting plugin for Kotlin. I have tested and it appears to work. It is very similar to my mining plugin. 

This has the same issue as the mining plugin when it comes to disappearing objects. But this time it is trees and stumps. 

Other than that, it looks like it all works. I would love some feed back!

Thank you,
Trevor


tlf30:
There also apperars to be an issue that depending on where you are standing to the tree, woodcutting is never started. The activity is never run.

tlf30:
I will rebase this and set it up with a async action.

tlf30:
This is rebased. I have not setup async yet

tlf30:
Sorry, something went really wrong on this rebase, I am not sure why the branched merged into itself...

PR #355, opened 2017-09-17
Added woodcutting plugin for Kotlin

tlf30:
OK, this is the new pr for wood cutting. The old branch imploded on a rebase. 

I need to migrate this to use async actions before it is ready to be merged.

### DIFF
--- a/game/plugin/skills/woodcutting/build.gradle
+++ b/game/plugin/skills/woodcutting/build.gradle
@@ -2,7 +2,7 @@ plugin {
     name = "woodcutting_skill"
     packageName = "org.apollo.game.plugin.skills.woodcutting"
     authors = [
-			"tlf30"
+        "tlf30"
     ]
     dependencies = ["api"]
 }
--- a/game/plugin/skills/woodcutting/build.gradle
+++ b/game/plugin/skills/woodcutting/build.gradle
@@ -1,0 +1,7 @@
+plugin {
+    name = "woodcutting_skill"
+    packageName = "org.apollo.game.plugin.skills.woodcutting"
+    authors = [
+			"tlf30"
+    ]
+}
--- a/game/plugin/skills/woodcutting/build.gradle
+++ b/game/plugin/skills/woodcutting/build.gradle
@@ -4,4 +4,5 @@ plugin {
     authors = [
 			"tlf30"
     ]
+    dependencies = ["api"]
 }
--- a/game/plugin/skills/woodcutting/meta.toml
+++ b/game/plugin/skills/woodcutting/meta.toml
@@ -1,0 +1,9 @@
+name = "woodcutting-skill"
+package = "org.apollo.game.plugin.skills.woodcutting"
+authors = [
+    "tlf30"
+ ]
+
+[config]
+srcDir = "src/"
+testDir = "test/"
--- a/game/plugin/skills/woodcutting/meta.toml
+++ b/game/plugin/skills/woodcutting/meta.toml
@@ -1,9 +1,0 @@
-name = "woodcutting-skill"
-package = "org.apollo.game.plugin.skills.woodcutting"
-authors = [
-    "tlf30"
- ]
-
-[config]
-srcDir = "src/"
-testDir = "test/"
--- a/game/plugin/skills/woodcutting/src/axe.kt
+++ b/game/plugin/skills/woodcutting/src/axe.kt
@@ -1,4 +1,4 @@
-package org.apollo.game.plugin.skills.mining
+package org.apollo.game.plugin.skills.woodcutting
 
 import org.apollo.game.model.Animation;
 
--- a/game/plugin/skills/woodcutting/src/axe.kt
+++ b/game/plugin/skills/woodcutting/src/axe.kt
@@ -2,7 +2,7 @@ package org.apollo.game.plugin.skills.woodcutting
 
 import org.apollo.game.model.Animation;
 
-//Animation IDs: https://www.rune-server.ee/runescape-development/rs2-client/configuration/272373-emote-gfx-id-list.html
+//Animation IDs thanks to Deadly A G S at rune-server.ee
 enum class Axe(val id: Int, val level: Int, val animation: Animation, val pulses: Int) {
     RUNE(1359, 41, Animation(867), 3),
     ADAMANT(1357, 31, Animation(869), 4),
@@ -10,14 +10,14 @@ enum class Axe(val id: Int, val level: Int, val animation: Animation, val pulses
     BLACK(1361, 11, Animation(873), 6),
     STEEL(1353, 6, Animation(875), 6),
     IRON(1349, 1, Animation(877), 7),
-    BRONZE(1351, 1, Animation(879), 8)
+    BRONZE(1351, 1, Animation(879), 8);
+
+    companion object {
+        private val AXES = Axe.values()
+        fun getAxes(): Array<Axe> {
+            return AXES
+        }
+    }
 }
 
-val AXES = Axe.values()
-
-fun getAxes(): Array<Axe> {
-    return AXES
-}
-
-fun lookupPickaxe(id: Int): Axe? = AXES.find { it.id == id }
 
--- a/game/plugin/skills/woodcutting/src/axe.kt
+++ b/game/plugin/skills/woodcutting/src/axe.kt
@@ -1,0 +1,23 @@
+package org.apollo.game.plugin.skills.mining
+
+import org.apollo.game.model.Animation;
+
+//Animation IDs: https://www.rune-server.ee/runescape-development/rs2-client/configuration/272373-emote-gfx-id-list.html
+enum class Axe(val id: Int, val level: Int, val animation: Animation, val pulses: Int) {
+    RUNE(1359, 41, Animation(867), 3),
+    ADAMANT(1357, 31, Animation(869), 4),
+    MITHRIL(1355, 21, Animation(871), 5),
+    BLACK(1361, 11, Animation(873), 6),
+    STEEL(1353, 6, Animation(875), 6),
+    IRON(1349, 1, Animation(877), 7),
+    BRONZE(1351, 1, Animation(879), 8)
+}
+
+val AXES = Axe.values()
+
+fun getAxes(): Array<Axe> {
+    return AXES
+}
+
+fun lookupPickaxe(id: Int): Axe? = AXES.find { it.id == id }
+
--- a/game/plugin/skills/woodcutting/src/wood.kt
+++ b/game/plugin/skills/woodcutting/src/wood.kt
@@ -1,0 +1,85 @@
+package org.apollo.game.plugin.skills.mining
+
+
+/**
+ * values thanks to: http://oldschoolrunescape.wikia.com/wiki/Woodcutting
+ */
+enum class Wood(val id: Int, val objects: IntArray, val stump: Int, val level: Int, val exp: Double, val chance: Double) {
+    NORMAL(1511, NORMAL_OBJECTS, NORMAL_STUMP, 1, 25.0, 100.0),
+    ACHEY(2862, ACHEY_OBJECTS, ACHEY_STUMP, 1, 25.0, 100.0),
+    OAK(1521, OAK_OBJECTS, OAK_STUMP, 15, 37.5, 0.125),
+    WILLOW(1519, WILLOW_OBJECTS, WILLOW_STUMP, 30, 67.5, 0.125),
+    TEAK(6333, TEAK_OBJECTS, TEAK_STUMP, 35, 85.0, 0.125),
+    MAPLE(1517, MAPLE_OBJECTS, MAPLE_STUMP, 45, 100.0, 0.125),
+    MAHOGANY(6332, MAHOGANY_OBJECTS, MAHOGANY_STUMP, 50, 125.0,0.125),
+    YEW(1515, YEW_OBJECTS, YEW_STUMP, 60, 175.0, 0.125),
+    MAGIC(1513, MAGIC_OBJECTS, MAGIC_STUMP, 75, 250.0, 0.125),
+}
+
+val WOOD = Wood.values()
+
+fun lookupWood(id: Int): Wood? =WOOD.find { it.id == id }
+
+fun lookupTree(id: Int): Wood? {
+    for (wood in WOOD) {
+        for (tree in wood.objects) {
+            if (tree == id) {
+                return wood
+            }
+        }
+    }
+    return null
+}
+
+val NORMAL_STUMP = 1342
+val ACHEY_STUMP = 3371
+val OAK_STUMP = 1342
+val WILLOW_STUMP = 1342
+val TEAK_STUMP = 1342
+val MAPLE_STUMP = 1342
+val MAHOGANY_STUMP = 1342
+val YEW_STUMP = 1342
+val MAGIC_STUMP = 1324
+
+val NORMAL_OBJECTS = intArrayOf(
+        1276, 1277, 1278, 1279, 1280, 1282, 1283, 1284, 1285, 1285, 1286, 1289, 1290, 1291, 1315,
+        1316, 1318, 1330, 1331, 1332, 1365, 1383, 1384, 2409, 3033, 3034, 3035, 3036, 3881, 3882,
+        3883, 5902, 5903, 5904, 10041
+)
+
+val ACHEY_OBJECTS = intArrayOf(
+        2023
+)
+
+val OAK_OBJECTS = intArrayOf(
+        1281, 3037
+)
+
+val WILLOW_OBJECTS = intArrayOf(
+        5551, 5552, 5553
+)
+
+val TEAK_OBJECTS = intArrayOf(
+        9036
+)
+
+
+val MAPLE_OBJECTS = intArrayOf(
+        1307, 4674
+)
+
+val MAHOGANY_OBJECTS = intArrayOf(
+        9034
+)
+
+
+val YEW_OBJECTS = intArrayOf(
+        1309
+)
+
+val MAGIC_OBJECTS = intArrayOf(
+        1292, 1306
+)
+
+
+
--- a/game/plugin/skills/woodcutting/src/wood.kt
+++ b/game/plugin/skills/woodcutting/src/wood.kt
@@ -1,4 +1,4 @@
-package org.apollo.game.plugin.skills.mining
+package org.apollo.game.plugin.skills.woodcutting
 
 
 /**
--- a/game/plugin/skills/woodcutting/src/wood.kt
+++ b/game/plugin/skills/woodcutting/src/wood.kt
@@ -1,85 +1,72 @@
 package org.apollo.game.plugin.skills.woodcutting
 
+private val NORMAL_STUMP = 1342
+private val ACHEY_STUMP = 3371
+private val OAK_STUMP = 1342
+private val WILLOW_STUMP = 1342
+private val TEAK_STUMP = 1342
+private val MAPLE_STUMP = 1342
+private val MAHOGANY_STUMP = 1342
+private val YEW_STUMP = 1342
+private val MAGIC_STUMP = 1324
+
+private val NORMAL_OBJECTS = hashSetOf(
+        1276, 1277, 1278, 1279, 1280, 1282, 1283, 1284, 1285, 1285, 1286, 1289, 1290, 1291, 1315,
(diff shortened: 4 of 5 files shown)


tlf30:
I have added async action. Now I need to test.

tlf30:
Wood cutting is ready now. There are only two issues with it: 
1. If the attack level is not high enough for the player to use the axe, there is no error message.
2. For some reason sometimes the action does not see the player as close enough. I am wondering if size should be 2.

game/plugin/skills/woodcutting/src/axe.kt
```diff
@@ -0,0 +1,23 @@
+package org.apollo.game.plugin.skills.woodcutting
+
+import org.apollo.game.model.Animation;
+
+//Animation IDs: https://www.rune-server.ee/runescape-development/rs2-client/configuration/272373-emote-gfx-id-list.html
```
Major-: Could you remove the link and just give the users name instead? Thanks!

game/plugin/skills/woodcutting/src/wood.kt
```diff
@@ -0,0 +1,87 @@
...
+val MAGIC_OBJECTS = intArrayOf(
+        1292, 1306
+)
+
+
+
+/**
+ * values thanks to: http://oldschoolrunescape.wikia.com/wiki/Woodcutting
+ */
+enum class Wood(val id: Int, val objects: IntArray, val stump: Int, val level: Int, val exp: Double, val chance: Double) {
```
Major-: Can we rename this to `Tree`?

game/plugin/skills/woodcutting/src/wood.kt
```diff
@@ -0,0 +1,87 @@
...
+val ACHEY_STUMP = 3371
+val OAK_STUMP = 1342
+val WILLOW_STUMP = 1342
+val TEAK_STUMP = 1342
+val MAPLE_STUMP = 1342
+val MAHOGANY_STUMP = 1342
+val YEW_STUMP = 1342
+val MAGIC_STUMP = 1324
+
+val NORMAL_OBJECTS = intArrayOf(
```
Major-: These should all be HashSets (`hashSetOf`) so we can get constant-time searches

game/plugin/skills/woodcutting/meta.toml
```diff
@@ -0,0 +1,9 @@
+name = "woodcutting-skill"
```
Major-: Now we have the `build.gradle` files we can delete the `meta.toml`s, so this can go

game/plugin/skills/woodcutting/build.gradle
```diff
@@ -0,0 +1,8 @@
+plugin {
+    name = "woodcutting_skill"
+    packageName = "org.apollo.game.plugin.skills.woodcutting"
+    authors = [
+			"tlf30"
```
Major-: Could you fix the indentation on this line (should be 8 spaces), thanks!

game/plugin/skills/woodcutting/src/axe.kt
```diff
@@ -0,0 +1,23 @@
...
+    MITHRIL(1355, 21, Animation(871), 5),
+    BLACK(1361, 11, Animation(873), 6),
+    STEEL(1353, 6, Animation(875), 6),
+    IRON(1349, 1, Animation(877), 7),
+    BRONZE(1351, 1, Animation(879), 8)
+}
+
+val AXES = enumValues<Axe>()
+
+fun getAxes(): Array<Axe> {
```
Major-: Is this function used anywhere?

game/plugin/skills/woodcutting/src/axe.kt
```diff
@@ -0,0 +1,23 @@
...
+    RUNE(1359, 41, Animation(867), 3),
+    ADAMANT(1357, 31, Animation(869), 4),
+    MITHRIL(1355, 21, Animation(871), 5),
+    BLACK(1361, 11, Animation(873), 6),
+    STEEL(1353, 6, Animation(875), 6),
+    IRON(1349, 1, Animation(877), 7),
+    BRONZE(1351, 1, Animation(879), 8)
+}
+
+val AXES = enumValues<Axe>()
```
Major-: Where's `enumValues` coming from?

game/plugin/skills/woodcutting/src/wood.kt
```diff
@@ -0,0 +1,87 @@
...
+    YEW(1515, YEW_OBJECTS, YEW_STUMP, 60, 175.0, 0.125),
+    MAGIC(1513, MAGIC_OBJECTS, MAGIC_STUMP, 75, 250.0, 0.125),
+}
+
+val WOOD = enumValues<Wood>()
+
+fun lookupWood(id: Int): Wood? = WOOD.find { it.id == id }
+
+fun lookupTree(id: Int): Wood? {
+    for (wood in enumValues<Wood>()) {
```
Major-: Should be able to replace this entire function with:

```kotlin
fun lookupTree(id: Int): Wood? = WOOD.find { it.objects.contains(id) }
```

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,117 @@
...
+import org.apollo.game.plugins.api.woodcutting
+import org.apollo.game.plugins.api.skills
+import org.apollo.game.scheduling.ScheduledTask
+import java.util.*
+
+class WoodcuttingTarget(val objectId: Int, val position: Position, val wood: Wood) {
+
+    fun getObject(world: World): Optional<GameObject> {
+        val region = world.regionRepository.fromPosition(position)
+        val obj = region.findObject(position, objectId)
```
Major-: I think this variable is pointless? Can just return the function result here.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,117 @@
...
+
+        return obj
+    }
+
+    fun isSuccessful(mob: Player): Boolean {
```
Major-: I think this must be influenced by the `Player`'s wc level in some way - if this is the case, can we add a TODO?

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        /**
+         * Starts a [WoodcuttingAction] for the specified [Player], terminating the [Message] that triggered it.
+         */
+        fun start(message: ObjectActionMessage, player: Player, wood: Wood) {
+            val axe = axeFor(player)
+            if (axe != null) {
+                val action = WoodcuttingAction(player, axe, WoodcuttingTarget(message.id, message.position, wood))
+                System.out.println("Starting action")
```
Major-: Can you remove the print statements please (and btw in kotlin this could just be `println(msg_here)`)

game/plugin/skills/woodcutting/src/axe.kt
garyttierney: My code.  I was seeing some interesting issues iterating over Axe.values().  Might have been local only.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,117 @@
...
+        if (mob.inventory.add(target.wood.id)) {
+            //TODO: Use lookup from utils once it has a lookup function for IDs
+            val woodName = ItemDefinition.lookup(target.wood.id).name.toLowerCase();
+            mob.sendMessage("You managed to cut some $woodName.")
+            mob.skillSet.addExperience(Skill.WOODCUTTING, target.wood.exp)
+        }
+
+        if (target.isSuccessful(mob)) {
+            //respawn time: http://runescape.wikia.com/wiki/Trees
+            val respawn = ((30 * 1000) / 600) + ((rand(150) * 1000) / 600) // between 30 sec and 3 min respawm
```
Major-:
```kotlin
val respawn = TimeUnit.SECONDS.toMillis(30 + rand(150)) / GameConstants.PULSE_DELAY
```

using `java.util.concurrent.TimeUnit` and `org.apollo.game.GameConstants`.
Might also be nice to put the `30` into its own constant - say, `MINIMUM_RESPAWN_TIME`.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,117 @@
...
+            mob.world.expireObject(obj.get(), target.wood.stump, respawn)
+            stop()
+        }
+    }
+}
+
+on { ObjectActionMessage::class }
```
Major-: Can you reformat this - should be 4 spaces for an indentation level, not 8 (with the latest `.editorconfig` if you have IDEA configured to follow it, this should be done automatically). Thanks!

Major-:
Good work, thanks a lot! Left some mostly small comments, but do try and make sure to select the appropriate data structure even in plugin code - if you need quick `contains` searches and don't care about ordering, a `HashSet` is definitely the way to go

game/plugin/skills/woodcutting/src/axe.kt
tlf30: No, I was having the same issue. So I saw your approach and that seemed to help. I was getting a static initialization error.

tlf30: So, if I move the vals that the enum uses to above the enum, then I get this:
```
SEVERE: Exception occurred during pulse!
java.lang.ExceptionInInitializerError
	at org.apollo.game.plugin.skills.woodcutting.WoodKt.<clinit>(wood.kt:22)
	at Woodcutting_plugin$3.invoke(woodcutting.plugin.kts:112)
	at Woodcutting_plugin$3.invoke(woodcutting.plugin.kts:16)
	at org.apollo.game.plugin.kotlin.KotlinPlayerHandlerProxyTrait$DefaultImpls.handleProxy(KotlinPluginScript.kt:81)
	at org.apollo.game.plugin.kotlin.KotlinMessageHandler.handleProxy(KotlinPluginScript.kt:97)
	at org.apollo.game.plugin.kotlin.KotlinMessageHandler.handleProxy(KotlinPluginScript.kt:97)
	at org.apollo.game.plugin.kotlin.KotlinMessageHandler.handle(KotlinPluginScript.kt:103)
	at org.apollo.game.message.handler.MessageHandlerChain.notify(MessageHandlerChain.java:57)
	at org.apollo.game.message.handler.MessageHandlerChainSet.notify(MessageHandlerChainSet.java:45)
	at org.apollo.game.session.GameSession.handlePendingMessages(GameSession.java:96)
	at org.apollo.game.service.GameService.pulse(GameService.java:162)
	at org.apollo.game.GamePulseHandler.run(GamePulseHandler.java:37)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.IllegalArgumentException: Parameter specified as non-null is null: method org.apollo.game.plugin.skills.woodcutting.Wood.<init>, parameter objects
	at org.apollo.game.plugin.skills.woodcutting.Wood.<init>(wood.kt)
	at org.apollo.game.plugin.skills.woodcutting.Wood.<clinit>(wood.kt:11)
	... 19 more


```

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
tlf30: This is the correct behavior. The player gets logs faster from a tree as his level goes up, but the tree always has the same chance at being cut down. Some trees will always get cut down and only yield one logs such as NORMAL, but other such as OAK will usually yield multiple logs before being cut down. http://oldschoolrunescape.wikia.com/wiki/Woodcutting#Trivia

Major-: Sorry, I thought this function was for gathering logs. Could we rename it to `fellsTree` or something?

game/plugin/skills/woodcutting/src/axe.kt
tlf30: Yes, in the lookup for which axe to use. That way if the implementation of storing the axes changes in the future, there will be less code to rewrite.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
tlf30: Done. Yeah, I need to get my idea set to use that. Idea is not happy for me right now.

tlf30:
So here is my question. The EquipItemHandler prevents the player from equipping an axe that he is not high enough level to equip. But the woodcutting action does not check for attack level. But somewhere there is a piece of code preventing the player from cutting down the tree until he has a high enough combat level to use the axe on a tree. Does anyone know where that code is? If so, can we have it send a message to the player about why he cannot cut down the tree.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
tlf30: Yeah, I already renamed it,

Promises:
You can always use a higher level axes in your inventory without having the attack lvl.

tlf30:
@Promises, thank you for the info. Now I need to figure out what is keeping the player from performing the action without the attack level

tlf30:
So, I am completely wrong. The issue is not related to attack, it is that rune axes are not cutting down normal trees. I will look into this.

tlf30:
I fixed it, the issue had nothing to do with the axe. It was because the code had trees as a 1x1 tile when in fact they are 2x2 tiles.

game/plugin/skills/woodcutting/src/wood.kt
```diff
@@ -0,0 +1,74 @@
...
+    MAPLE(1517, MAPLE_OBJECTS, MAPLE_STUMP, 45, 100.0, 0.125),
+    MAHOGANY(6332, MAHOGANY_OBJECTS, MAHOGANY_STUMP, 50, 125.0, 0.125),
+    YEW(1515, YEW_OBJECTS, YEW_STUMP, 60, 175.0, 0.125),
+    MAGIC(1513, MAGIC_OBJECTS, MAGIC_STUMP, 75, 250.0, 0.125),
+}
+
+val TREES = Tree.values()
+
+fun lookupTree(id: Int): Tree? = TREES.find { it.objects.contains(id) }
+
```
Major-: Can you leave either zero or one new line at the bottom of the file please

game/plugin/skills/woodcutting/src/wood.kt
```diff
@@ -0,0 +1,74 @@
...
+    TEAK(6333, TEAK_OBJECTS, TEAK_STUMP, 35, 85.0, 0.125),
+    MAPLE(1517, MAPLE_OBJECTS, MAPLE_STUMP, 45, 100.0, 0.125),
+    MAHOGANY(6332, MAHOGANY_OBJECTS, MAHOGANY_STUMP, 50, 125.0, 0.125),
+    YEW(1515, YEW_OBJECTS, YEW_STUMP, 60, 175.0, 0.125),
+    MAGIC(1513, MAGIC_OBJECTS, MAGIC_STUMP, 75, 250.0, 0.125),
+}
+
+val TREES = Tree.values()
+
+fun lookupTree(id: Int): Tree? = TREES.find { it.objects.contains(id) }
```
Major-: For lookup we're really just trying to map from object_id to Tree:

```kotlin
private val TREES = Trees.values().flatMap { tree -> tree.objects.map { Pair(it, tree) } }.toMap()

fun lookup(id: Int): Tree? = TREES[id]
```

If you could put that in a companion object inside `Tree` as well that'd be great (for enum classes the companion objects need to be located beneath all the enumerators, similar to Java).

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,122 @@
...
+     */
+    fun isCutDown(mob: Player): Boolean {
+        return rand(100) <= tree.chance * 100
+    }
+}
+
+class WoodcuttingAction(val player: Player, val tool: Axe, val target: WoodcuttingTarget) : AsyncDistancedAction<Player>(PULSES, true, player, target.position, TREE_SIZE) {
+
+    companion object {
+        private val PULSES = 0
```
Major-: Rename this to `DELAY` please.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,122 @@
...
+
+        //Check that the object exists in the world
+        val obj = target.getObject(mob.world)
+        if (!obj.isPresent) {
+            stop()
+            return
+        }
+
+        //Check that the player had room in their inventory
+        if (mob.inventory.freeSlots() == 0) {
```
Major-: Apollo already handles capacity checking and notification (i.e. `forceCapacityExceeded()`) when you attempt to add items. However, in RS it actually checks if you have inventory space at the start of the action (before you play the animation), so basically we want to move this check up to the very top of this function.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,122 @@
...
+        /**
+         * Starts a [WoodcuttingAction] for the specified [Player], terminating the [Message] that triggered it.
+         */
+        fun start(message: ObjectActionMessage, player: Player, wood: Tree) {
+            val axe = axeFor(player)
+            if (axe != null) {
+                val action = WoodcuttingAction(player, axe, WoodcuttingTarget(message.id, message.position, wood))
+                player.startAction(action)
+            } else {
+                player.sendMessage("You do not have a axe for which you have the level to use.")
```
Major-: should be "an axe"

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
```diff
@@ -0,0 +1,122 @@
...
+
+        //Check that the player had room in their inventory
+        if (mob.inventory.freeSlots() == 0) {
+            mob.inventory.forceCapacityExceeded()
+            stop()
+            return
+        }
+        if (mob.inventory.add(target.tree.id)) {
+            //TODO: Use lookup from utils once it has a lookup function for IDs
+            val woodName = ItemDefinition.lookup(target.tree.id).name.toLowerCase();
```
Major-: `logName`

game/plugin/skills/mining/build.gradle
```diff
@@ -2,12 +2,12 @@ plugin {
     name = "mining-skill"
```
Major-: Can you remove the change to the mining build.gradle from this plugin please? The indentation does look off but each PR should ideally be a set of related changes.

tlf30: Yeah, sorry. Fighting idea.

game/plugin/skills/woodcutting/src/wood.kt
tlf30: That is a good idea!.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
tlf30: Good to know. I will move this to the start in the companion object.

game/plugin/skills/woodcutting/src/woodcutting.plugin.kts
tlf30: Ops, grammar error. Sorry.

tlf30:
How does that look?

tlf30:
This builds locally for me

gruckion:
@tfl30, what about cutting a small tree while being one space away. Since all trees are not 2x2, will this present a problem?

Major-:
@gruckion Yeah this is a known problem with DistancedAction - because it uses a range (distance) check rather than actually checking the tiles you can interact with the object from, you need to specify an N range if the object is NxN. DistancedAction is going to be completely replaced with an implementation that does this properly Soon (tm), which will solve this.

In other words, its a problem with core not this plugin.

PR #356, opened 2017-09-17
Support for doors in Kotlin

arinhouck:
Hey everyone,

I just jumped on board. I hadn't done any RSPS programming since 2009-2012 range. I remembered Apollo as a project but I was never experienced enough back then to contribute.

However fast forward to now with much more programming experience, and I am always about advancing my programming skills in new ways. So I decided to give it a go on doors.

I followed a similar pattern to the functionality of the existing door code. I tried to clean it up and explain it better some as it the code was rather hard to interpret through the previous DoorUtil class, especially within Ruby. See for yourself.

I still need to write test cases but I wanted to get opinions on how to go about that and also if anyone has suggestions/feedback on the code I am presenting now.

Cheers.


tlf30:
Nice, good job!

game/plugin/navigation/door/src/door.plugin.kts
```diff
@@ -0,0 +1,212 @@
...
+import org.apollo.game.model.Position
+import org.apollo.game.model.World
+import org.apollo.game.model.event.PlayerEvent
+import org.apollo.game.message.impl.ObjectActionMessage
+import org.apollo.net.message.Message
+
+enum class Orientation(val value: Int) {
+    WEST(0), NORTH(1), EAST(2), SOUTH(3);
+    companion object {
+        fun from(findValue: Int): Orientation = Orientation.values().first { it.value == findValue }
```
Major-: I think `findValue` would be more accurately named `ordinal`.

game/plugin/navigation/door/src/door.plugin.kts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+class Door(private val gameObject: GameObject) {
+
+    companion object {
+
+        /**
+         * Mapping to change orientation of a left hinged door
+         */
+        val LEFT_HINGE_ORIENTATION: HashMap<Orientation, Orientation> = hashMapOf(
```
Major-: Are these type annotations required?

game/plugin/navigation/door/src/door.plugin.kts
```diff
@@ -0,0 +1,212 @@
...
+        /**
+         * Find a given door in the world
+         * @param world The [World] the door lives in
+         * @param position The [Position] of the door
+         * @param objectId The [GameObject] id of the door
+         */
+        fun find(world: World, position: Position, objectId: Int): Door? {
+            val objects: Set<GameObject> = world.regionRepository.fromPosition(position).getEntities(position, EntityType.DYNAMIC_OBJECT, EntityType.STATIC_OBJECT)
+            val gameObject: GameObject? = objects.find { it.id == objectId }
+            return if(gameObject == null) {
```
Major-: Can you add spaces after `if` please (line 182 has the same problem)

game/plugin/navigation/door/src/door.plugin.kts
```diff
@@ -0,0 +1,212 @@
...
+
+        /**
+         * Find a given door in the world
+         * @param world The [World] the door lives in
+         * @param position The [Position] of the door
+         * @param objectId The [GameObject] id of the door
+         */
+        fun find(world: World, position: Position, objectId: Int): Door? {
+            val objects: Set<GameObject> = world.regionRepository.fromPosition(position).getEntities(position, EntityType.DYNAMIC_OBJECT, EntityType.STATIC_OBJECT)
+            val gameObject: GameObject? = objects.find { it.id == objectId }
```
Major-: You can express this more concisely using the [elvis operator](https://kotlinlang.org/docs/reference/null-safety.html#elvis-operator):
```kotlin
val gameObject: GameObject = objects.find { it.id == objectId } ?: return null
return Door(gameObject)
```

game/plugin/navigation/door/src/door.plugin.kts
```diff
@@ -0,0 +1,212 @@
...
+         */
+        fun start(message: Message, player: Player, door: Door, position: Position) {
+            player.startAction(OpenDoorAction(player, door, position))
+            message.terminate()
+        }
+
+    }
+
+    override fun executeAction() {
+        if(player.world.submit(OpenDoorEvent(player))) {
```
Major-: Can you add a space after `if` please

game/plugin/navigation/door/src/door.plugin.kts
garyttierney: This can be reduced to

```kotlin
import org.apollo.game.plugins.api.*

val gameObject = region.findObject(position, id)
```

Since https://github.com/apollo-rsps/apollo/commit/cf0e3331e0fec57169f0fa0a6dd7be110e4e694f

Major-: Nice!

game/plugin/navigation/door/src/door.plugin.kts
```diff
@@ -0,0 +1,212 @@
...
+
+    }
+
+    /**
+     * Calculates the position to move the door based on orientation
+     */
+    private fun movePosition(): Position {
+        val position = gameObject.position
+        return when(Orientation.from(gameObject.orientation)) {
+            Orientation.WEST -> Position(position.x - 1, position.y, position.height)
```
garyttierney:
```kotlin 
position.step(1, direction)
```

game/plugin/navigation/door/src/door.plugin.kts
garyttierney: Please use `Direction` and `Direction.WNES` instead.

game/plugin/navigation/door/src/door.plugin.kts
arinhouck: When storing region as a variable, are you suggesting
`Region region = world.regionRepository.fromPosition(position)`
As I am having issues defining this line. @garyttierney

garyttierney: Yes,

```kotlin
val region = world.regionRepository.fromPosition(position)
val object = region.findObject(position, id)
```

game/plugin/navigation/door/src/door.plugin.kts
arinhouck: Can you enlighten me on how I would get direction from gameObject orientation integer?

As currently, it is causing me issues for this refactoring. The following code getting orientation/direction by position from the gameObject is causing issues during removal of an entity. Not seeing how it is related. @garyttierney 
```
Direction.fromDeltas(position.x, position.y)
```

```
private fun movePosition(): Position {
        val position = gameObject.position
        System.out.println("DIRECTION" + Direction.fromDeltas(position.x, position.y).toInteger())
        return when(Direction.fromDeltas(position.x, position.y)) {
            Direction.WEST -> position.step(1, Direction.WEST)
            Direction.EAST -> position.step(1, Direction.EAST)
            Direction.NORTH -> position.step(1, Direction.NORTH)
            Direction.SOUTH -> position.step(1, Direction.SOUTH)
            else -> return position
        }

    }
```

garyttierney:
```kotlin
Direction.WNES[object.orientation]
```

That could be made clearer.

arinhouck:
I have refactored to use Direction, rebased with the latest, cleaned up syntax, and refactored the file to only have implement action inside the script plugin and the door class in a separate kotlin file.

The one last change I wasn't able to complete was using `findObject` as @garyttierney had suggested. I may be missing something but I tried the same snippet from the mining plugin using it and it will not compile. It doesn't recognize `findObject` as an instance method of Region type. Maybe I am missing something but it will not compile for me if someone could give it a try that would be great.

arinhouck:
Got it figured out thanks to @tlf30 mentioning that I needed to add a dependency in my build.gradle. Not sure why it required me to explicitly `import findObject`, possibly because it is not the root plugin file? However, the refactor is complete now.

game/plugin/navigation/door/meta.toml
```diff
@@ -0,0 +1,7 @@
+name = "Door"
```
Major-: We aren't using `meta.toml` anymore, can you delete this file and add the appropriate information to `build.gradle` instead please (you can look at any of the existing plugins for the format).

Also, *please* add the authors of the previous door plugin to the authors list.

game/plugin/navigation/door/src/door.kt
```diff
@@ -0,0 +1,180 @@
...
+
+            val toggledDoor = DynamicGameObject.createPublic(world, gameObject.id, position, gameObject.type, orientation)
+
+            // Renders new door
+            regionRepository.fromPosition(position).addEntity(toggledDoor)
+            toggledDoors.put(toggledDoor, gameObject)
+        } else {
+            // Removes toggled door
+            toggledDoors.remove(gameObject)
+            // Renders original door
```
Major-: Comments should explain why a code is doing something, not describe what it is doing, so you can delete the inline ones (the `//` ones) here. If everything you get from a comment you can get just from reading the code, you don't need the comment.

game/plugin/navigation/door/src/door.kt
```diff
@@ -0,0 +1,180 @@
...
+    override fun executeAction() {
+        if (player.world.submit(OpenDoorEvent(player))) {
+            player.turnTo(position)
+            door.toggle()
+        }
+        stop()
+    }
+
+    override fun equals(other: Any?): Boolean {
+        return other is OpenDoorAction && position == other.position
```
Major-: I think we also want `player == other.player` here

game/plugin/navigation/door/src/door.plugin.kts
```diff
@@ -0,0 +1,16 @@
...
+import org.apollo.plugin.navigation.door.Door
+import org.apollo.plugin.navigation.door.OpenDoorAction
+
+/**
+ * Hook into the [ObjectActionMessage] and listens for a supported door [GameObject]
+ */
+on { ObjectActionMessage::class }
+        .where { option == 1 }
+        .then {
+            val door: Door? = Door.find(it.world, position, id)
```
Major-: You can do the "if null then return" stuff on the assignment line, which is quite nice to avoid boilerplate further down:

```kotlin
val door = Door.find(it.world, position, id) ?: return@then
```

game/plugin/navigation/door/src/door.kt
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    override fun hashCode(): Int {
```
Major-:
```kotlin
override fun hashCode(): Int = Objects.hash(position, player)
```

game/plugin/navigation/door/src/door.kt
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+    }
+
+    /**
+     * Calculates the position to move the door based on orientation
+     */
+    private fun movePosition(): Position {
+        val position = gameObject.position
+        return when(Direction.WNES[gameObject.orientation]) {
```
Major-:
```kotlin
return position.step(1, Direction.WNES[gameObject.orientation])
```

Major-:
If you can address the changes I've noted we can merge this. Thanks a lot!

garyttierney:
> Not sure why it required me to explicitly import findObject, possibly because it is not the root plugin file? 

`findObject` is an extension method of `World`, so you need to import it to be able to call that method on `World` objects.  I hope to make this simpler in a later revision of the plugin system.

arinhouck:
See commit message above for more information on latest changes. Should be good to go now @Major- @garyttierney. Thanks guys for the guidance and help. Look forward to seeing this project thrive. I would love to hear more about what is going on (roadmap, milestones, priorities, etc) to get a better understanding of how I can contribute. In addition, the IRC isn't as inviting for newcomers as you can't see when the last discussion was, sharing code and screenshots is difficult (as you have to upload externally), and no way to contribute to conversions in the past. I think we could use another tool for better communication (eg. Slack, Discord). If that issue was addressed, the project could potentially gain more contributors and help existing contributors work together more effectively.

Major-:
Some really good work, thanks! Can definitely understand your perspective on irc, gary and I will talk about it - we both prefer irc but you also aren't wrong about things like conversation history for newcomers.

PR #358, opened 2017-09-18
Add logging to LoginDecoder for session rejections

### LINKED ISSUE #322: Log information when a login session is rejected
We should log debug information in `LoginDecoder`, when sessions are rejected:

```java
writeResponseCode(ctx, LoginConstants.STATUS_LOGIN_SERVER_REJECTED_SESSION);
```

Information to include:
- The stage that caused the rejection (e.g. RSA magic mismatch, seed mismatch)
- more?

Username cant be included because its only decoded after this (and will probably be garbage data if we need to reject a session anyway)

kylestev:
Fixes #322

Not sure about the logging level here -- open to suggestions.

I was manually setting breakpoints to debug a 377 client that had the modulus and public key params mixed up. Less than ideal 😅

### DIFF
--- a/net/src/main/org/apollo/net/codec/login/LoginDecoder.java
+++ b/net/src/main/org/apollo/net/codec/login/LoginDecoder.java
@@ -9,6 +9,7 @@ import java.math.BigInteger;
 import java.net.InetSocketAddress;
 import java.security.SecureRandom;
 import java.util.List;
+import java.util.logging.Logger;
 
 import org.apollo.cache.FileSystemConstants;
 import org.apollo.net.NetworkConstants;
@@ -26,6 +27,11 @@ import com.google.common.net.InetAddresses;
  * @author Graham
  */
 public final class LoginDecoder extends StatefulFrameDecoder<LoginDecoderState> {
+
+	/**
+	 * The logger for this class.
+	 */
+	private static final Logger logger = Logger.getLogger(LoginDecoder.class.getName());
 
 	/**
 	 * The secure random number generator.
@@ -110,6 +116,7 @@ public final class LoginDecoder extends StatefulFrameDecoder<LoginDecoderState> 
 			int type = buffer.readUnsignedByte();
 
 			if (type != LoginConstants.TYPE_STANDARD && type != LoginConstants.TYPE_RECONNECTION) {
+				logger.fine("Failed to decode login header.");
 				writeResponseCode(ctx, LoginConstants.STATUS_LOGIN_SERVER_REJECTED_SESSION);
 				return;
 			}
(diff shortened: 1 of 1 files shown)


Major-:
@kylestev nice one, thanks!

@garyttierney do we want to log more robustly than this (db?)? All of these only (or should only) occur with modified clients

tlf30:
For the login fail, I think it would be nice to log the username of the failed login. @Major- what do you think?

kylestev:
@tlf30 qutoing #322 

>Username cant be included because its only decoded after this (and will probably be garbage data if we need to reject a session anyway)

tlf30:
@kylestev, I was thinking about when you logged 
```
				logger.warning("Username or password did not pass validation.");
  				writeResponseCode(ctx, LoginConstants.STATUS_INVALID_CREDENTIALS);
```
At this point in the code if the checks about pass, most likely the username should be good data.

kylestev:
Ah! I see what you're saying. Sure I think that could be helpful.

kylestev:
Addressed in d958435

net/src/main/org/apollo/net/codec/login/LoginDecoder.java
```diff
@@ -180,6 +191,7 @@ private void decodePayload(ChannelHandlerContext ctx, ByteBuf buffer, List<Objec
 			String hostAddress = InetAddresses.toAddrString(socketAddress.getAddress());
 
 			if (password.length() < 6 || password.length() > 20 || username.isEmpty() || username.length() > 12) {
+				logger.warning("Username ('" + username + "') or password (length: " + password.length() + ") did not pass validation.");
```
kylestev: Could drop the password length here but it might be helpful.

kylestev: Actually, this could be a potential cause for security concerns (password length being leaked on a typo'd username, for example). I'll remove it for now and just leave the username.

kylestev:
Okay removed the password length and just left the username in that specific log entry.

tlf30:
Sweet!
Thank you for adding the logging!

garyttierney:
> @garyttierney do we want to log more robustly than this (db?)? All of these only (or should only) occur with modified clients

I don't think so.  This suffices I think.

> Okay removed the password length and just left the username in that specific log entry.

I'd argue that logging a message in the condition that a password is too short is fine.

> `logger.warning("Username or password did not pass validation.");`

Please drop the logging level to `DEBUG`.  If someone is hammering my server with bad credentials I don't want my log filling up with warnings.

garyttierney:
As a side note: I'd prefer if the only signed commits in Apollo were releases from either myself or @Major- (or other maintainers) to prevent confusion for people viewing those signatures.  Thanks!

kylestev:
>As a side note: I'd prefer if the only signed commits in Apollo were releases from either myself or @Major- (or other maintainers) to prevent confusion for people viewing those signatures. Thanks!

@garyttierney sure thing! Just ran `git config commit.gpgsign false` in my local repo will force push to my branch to remove the signed commit.

>Please drop the logging level to DEBUG. If someone is hammering my server with bad credentials I don't want my log filling up with warnings.

Went with `Level.FINE`. Good catch!

kylestev:
>I'd argue that logging a message in the condition that a password is too short is fine.

I can make this change too if you want. I'd probably split the username and password validation into two separate if statements.

garyttierney:
> Went with `Level.FINE`. Good catch!

`FINE` is good.  LGTM.

PR #348, opened 2017-09-10
Added shops plugin for Kotlin

tlf30:
I have a rough port of the existing shops plugin in Kotlin. 
Just as in the ruby version, this only has shops for Varrock. 

This is currently untested. 

Some things had to work quite a bit different in kotlin than in ruby. I believe that this implementation is better and more flexible. But I would love some feed back. 

Thanks, 
Trevor

PR #360, opened 2017-09-23
Add shops in kotlin

Major-:
Based on some great work from @tlf30 in #348, who ported the ruby plugin and the varrock shops. Changes since that include:

- Clean up the main shop code (which is from the crap ruby plugin that I wrote)
- Add a kotlin DSL
- Fix some bugs pointed out by Scu11 🤓

This also adds support for using the `lookup_{item, npc, object}` utility functions with an id appended to the string (which existed in the ruby version). Although this is utilised only as an internal implementation detail, we probably want to support this behaviour anyway.

## DSL

Example of the old ruby DSL:
```ruby
create_shop npcs: :aubury, name: "Aubury's Rune Shop.", items: [
        [:fire_rune, 5_000], [:water_rune, 5_000], [:air_rune, 5_000], [:earth_rune, 5_000],
        [:mind_rune, 5_000], [:body_rune, 5_000], [:chaos_rune, 250], [:death_rune, 250]
]
```

And the equivalent in the kotlin DSL:
```kotlin
shop("Aubury's Rune Shop.") {
    category("runes") {
        sell(5000) of {
            -"Earth"
            -"Water"
            -"Fire"
            -"Air"
            -"Mind"
            -"Body"
        }

        sell(250) of {
            -"Chaos"
            -"Death"
        }
    }
}
```

The benefits of the new DSL are particularly obvious for shops with many items (especially ones with very similar amounts or prefixes/suffixes).

## Further work

- Unit tests (there's a fairly large amount of testable components). Might do this myself and amend them to this PR.
- Shops for other locations (definitely an `easy` issue, although time consuming).
- Add a tutorial on the wiki on using the plugin (needs to list all the dsl configuation options, maybe I should do this).
- Investigate ways to integrate this work with:
    - Quest, minigame, and custom plugins (including: non-default shop interface ids, requirements for accessing the shop, purchase item prices that are not derived from their `value` stored in the resources).
    - Dialogue (hopefully relatively trivial, although dialogue sequences such as `dialogue -> open shop -> dialogue (on shop close)` would be more complicated).
- Add support for some special general stores (Bandit duty free, Martin Thwait's Lost and Found) that buy for high alchemy price instead of low. Might be nice to try and combine this with the custom buy price work listed above.
- Implement the [price fluctuation algorithm](http://oldschoolrunescape.wikia.com/wiki/General_store#Sell_Prices) used by the general stores (shops that buy `any` items). Probably an `easy` issue.

### DIFF
--- a/game/plugin/locations/varrock/build.gradle
+++ b/game/plugin/locations/varrock/build.gradle
@@ -1,10 +1,12 @@
 plugin {
-    name = "varrock_npc_spawns"
+    name = "varrock"
     packageName = "org.apollo.game.plugin.locations"
     authors = [
-        "Jesse W",
+            "Jesse W",
+            "Major",
+            "tlf30",
     ]
     dependencies = [
-        "entity:spawn",
+            "entity:spawn", "shops"
     ]
 }
--- a/game/plugin/locations/varrock/build.gradle
+++ b/game/plugin/locations/varrock/build.gradle
@@ -1,6 +1,6 @@
 plugin {
     name = "varrock"
-    packageName = "org.apollo.game.plugin.locations"
+    packageName = "org.apollo.game.plugin.locations.varrock"
     authors = [
             "Jesse W",
             "Major",
--- a/game/plugin/shops/src/shop.plugin.kts
+++ b/game/plugin/shops/src/shop.plugin.kts
@@ -1,0 +1,46 @@
+
+import org.apollo.game.message.handler.ItemVerificationHandler
+import org.apollo.game.message.impl.ItemActionMessage
+import org.apollo.game.message.impl.NpcActionMessage
+import org.apollo.game.model.entity.Mob
+import org.apollo.game.scheduling.ScheduledTask
+
+fun Mob.shop(): Shop? = SHOPS[definition.id]
+
+start {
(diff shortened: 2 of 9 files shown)


game/plugin/locations/varrock/build.gradle
```diff
@@ -1,10 +1,12 @@
 plugin {
-    name = "varrock_npc_spawns"
+    name = "varrock"
     packageName = "org.apollo.game.plugin.locations"
```
garyttierney: Might want to update this to `org.apollo.game.plugin.locations.varrock` so it doesn't conflict with the next shops.plugin.kts added for a different location.

game/plugin/shops/src/shop.plugin.kts
```diff
@@ -0,0 +1,46 @@
...
+        override fun execute() = SHOPS.values.distinct().forEach(Shop::restock)
+    })
+}
+
+on { NpcActionMessage::class }
+    .then {
+        val npc = it.world.npcRepository.get(index)
+        val shop = npc.shop() ?: return@then
+
+        if (shop.action == option) {
```
garyttierney: It'd be good if we intercepted this message and set "option" to a normalized string value.

PR #369, opened 2017-10-18
Added Gradle wrapper

paramvir-phagura:
Builds shouldn't be reliant on the user's Gradle installation. Consistency can be ensured across them via a wrapper.


PavanChohan:
Not sure why the Travis CI build failed, apparently it's due to:

`/home/travis/.travis/job_stages: line 236: ./gradlew: Permission denied`

garyttierney:
Hi, thanks for the contribution.

>Not sure why the Travis CI build failed, apparently it's due to:
>
>`/home/travis/.travis/job_stages: line 236: ./gradlew: Permission denied`

That file would need the executable bit set on its permissions mask.  i.e. `chmod +x ./gradlew`.

That said, I'm not eager to include binaries in the repository.  There's also the associated maintenance cost of updating the gradle wrapper properties when new versions of Gradle are released.  If we do require a specific version of Gradle then the wrapper can be downloaded easily using the `gradle wrapper` task and as far as I'm aware Apollo currently builds with Gradle versions 2.x through 4.x.

Are you seeing issues using your system version of Gradle that is solved with the 4.2.1 release in this patch?

PavanChohan:
@garyttierney Hey, appreciate the response. To answer your question, no there aren't. However I believe it'd be easier for users to build the project with one specific version as there are several differences among 2 through 4. Therefore newer features can be used in buildscripts (if ever needed). This also provides an opportunity for potentially making use of [Kotlin buildscripts](https://github.com/gradle/kotlin-dsl) and Java 9 modules.

As far as maintainability overhead is concerned, major Gradle releases [don't happen very often](https://gradle.org/releases/). Where I do see overhead though is having to support multiple versions without a wrapper. But of course there's the alternative to have users manually run a wrapper task (like you said).

However if this conflicts with one of this projects goals, feel free to close this pull request. :)

Major-:
Closing for now, but this is a good question to revisit at a later date

garyttierney:
Due to backwards compatibility problems with the latest minor releases of Gradle I'm re-opening this and adding the Gradle wrapper for 4.9.

PR #370, opened 2017-11-06
Add play sound message

tlf30:
I know that kotin-experiments is closed to new features. I am opening this for reference until kotlin-experiments is merged into master. 

~Trevor

### DIFF
--- a/game/src/main/java/org/apollo/game/message/impl/SendPlaySoundMessage.java
+++ b/game/src/main/java/org/apollo/game/message/impl/SendPlaySoundMessage.java
@@ -16,11 +16,13 @@ public final class SendPlaySoundMessage extends Message {
 
 	/**
 	 * The delay that the client should use before playing the sound.
+	 * Deplay is measured in milliseconds.
 	 */
 	private final int delay;
 
 	/**
 	 * The type of sound to play.
+	 * Usually 0.
 	 */
 	private final int type;
 
@@ -28,8 +30,8 @@ public final class SendPlaySoundMessage extends Message {
 	 * Creates a new send play sound message.
 	 *
 	 * @param id The id of the sound to play.
-	 * @param type The type of the sound to play.
-	 * @param delay The delay before the client plays the sound
+	 * @param type The type of the sound to play, usually 0.
+	 * @param delay The delay in milliseconds before the client plays the sound
 	 */
 	public SendPlaySoundMessage(int id, int type, int delay) {
 		this.id = id;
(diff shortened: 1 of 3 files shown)


game/src/main/java/org/apollo/game/message/impl/SendPlaySoundMessage.java
```diff
@@ -0,0 +1,67 @@
...
+	 * The type of sound to play.
+	 */
+	private final int type;
+
+	/**
+	 * Creates a new send play sound message.
+	 *
+	 * @param id The id of the sound to play.
+	 * @param type The type of the sound to play.
+	 * @param delay The delay before the client plays the sound
```
garyttierney: What unit is the delay measured in?

game/src/main/java/org/apollo/game/message/impl/SendPlaySoundMessage.java
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * The type of sound to play.
+	 */
+	private final int type;
+
+	/**
+	 * Creates a new send play sound message.
+	 *
+	 * @param id The id of the sound to play.
+	 * @param type The type of the sound to play.
```
garyttierney: Do we know what the different types are?

game/src/main/java/org/apollo/game/message/impl/SendPlaySoundMessage.java
tlf30: I believe ticks

game/src/main/java/org/apollo/game/message/impl/SendPlaySoundMessage.java
tlf30: So far all the sounds I have used are using type 0

garyttierney:
Could you update the documentation about the two things mentioned (delay in ticks, type usually 0).  Besides that LGTM.

tlf30:
@garyttierney I looked at the client I have a little. The delay is in milliseconds. The type I am not sure of though. It is passed to a function in the SoundTrack class called mix(int loop). I played several sounds doing a test of type 0, the of type 1. I could not hear any difference.

garyttierney:
Ok.  If the type is always 0 we should add a constructor that uses 0 as a default value.

Major-:
Aren't there two different frames to play (different types?) of sound

garyttierney:
There's music and there's sound effects IIRC.

tlf30:
If someone can confirm which of the two channels are which, i.e channel 1 is music and 2 is sound effects. That is an example, I have no evidence for that. Then, I will work on adding sounds once kotlin experiments is merged. I have a list of sound effects and what they go to.

tlf30:
~~OK, I believe that channel 0 is sound effects and channel 1 is music.~~
No idea, music is on its own channel

PR #372, opened 2017-11-12
Ground items kotlin

rmcmk:
- Needs tests
- Need to solve that eyesore in `Region`
- Would be cool to include picking up items and other basic functionality

### DIFF
--- a/game/plugin/entity/ground-item/src/ground_item_sync.kt
+++ b/game/plugin/entity/ground-item/src/ground_item_sync.kt
@@ -1,0 +1,72 @@
+import org.apollo.game.GameConstants
+import org.apollo.game.model.entity.GroundItem
+import org.apollo.game.scheduling.ScheduledTask
+
+/**
+ * A [ScheduledTask] that manages the globalization and expiration of [GroundItem]s.
+ */
+class GroundItemSynchronizationTask(private val groundItem: GroundItem) : ScheduledTask(DELAY, false) {
+
+    companion object {
+
+        /**
+         * The delay between executions of this task, in pulses.
+         */
+        const val DELAY = 1
+
+        /**
+         * The amount of time, in pulses, in which this [GroundItem] will be globally visible.
+         */
+        const val TRADABLE_TIME_UNTIL_GLOBAL = 60000 / GameConstants.PULSE_DELAY
+
+        /**
+         * The amount of time, in pulses, in which this [GroundItem] will expire and be removed from the [World].
+         */
+        const val UNTRADABLE_TIME_UNTIL_EXPIRE = 180000 / GameConstants.PULSE_DELAY
+
+        /**
+         * The amount of time, in pulses, in which this [GroundItem] will expire and be removed from the [World].
+         */
+        const val TIME_UNTIL_EXPIRE = 180000 / GameConstants.PULSE_DELAY
+    }
+
+    /**
+     * The amount of game pulses this [ScheduledTask] has been alive.
+     */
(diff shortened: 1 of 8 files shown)


garyttierney:
It'd be nice to get transactions in with this before kotlin-experiments is merged back to master.

garyttierney:
> Needs tests

You can use the existing mocks of the `World` and `Player` to register a task and manually advance it to the delays you've set. Then you can run the assertions that the item was globalized / deleted.  There should be an example in `BankTests`.

Major-:
>It'd be nice to get transactions in with this before kotlin-experiments is merged back to master.

Not sure I agree with this. kotlin-experiments is already massive, we were supposed to have feature-frozen it

tlf30:
I agree with @Major-, I love the work that @ryleykimmel has done here, but it would be better to put merge this once kotlin-experiments is merged. Kotlin-experiments branch is already going to be a huge merge. 
Again, I love the work that @ryleykimmel has done here, and it definitely needs to be merged at some point.

garyttierney:
That's fine.  I'm more concerned with this making it to master before the transactional item stuff and then requiring a rewrite.

ryleykimmel:
FWIW the item transaction code I wrote many years ago should be scrapped. I don't remember the plan I had in mind when creating this and feel it could be (re)designed in a better manner.

Major-:
Need to decide whether or not this should be merged or should actually be in core (i.e. written in Java)

game/plugin/entity/ground-item/src/ground_item_sync.kt
```diff
@@ -0,0 +1,72 @@
+import org.apollo.game.GameConstants
+import org.apollo.game.model.entity.GroundItem
+import org.apollo.game.scheduling.ScheduledTask
+
+/**
+ * A [ScheduledTask] that manages the globalization and expiration of [GroundItem]s.
+ */
+class GroundItemSynchronizationTask(private val groundItem: GroundItem) : ScheduledTask(DELAY, false) {
```
garyttierney, line 8: I think this could be better done as a single task that polls GroundItem's from a queue every tick.

garyttierney (approved):

game/plugin/entity/ground-item/src/ground_item_sync.kt
rmcmk: Mind explaining how that queue should work? I don't think to queue them makes much sense here as all this task is doing is keeping the ground items state in sync with the client at the appropriate time.

Major-: So there's one GroundItemSyncTask at all times and new ones are just added to its queue, rather than creating a new task for each item. Downside of this is we have to implement what is essentially scheduling logic in the task itself

rmcmk: Kind of defeats the purpose of the task scheduler, does it not?

Major-: Partially, yes; that is the downside

rmcmk:
I'm looking at revisiting this. Can we come to a consensus on whether this should be core functionality or left alone in Kotlin?

Personally I think it fits better in the core as this kind of thing isn't really fit for a plugin as it is core functionality to how ground items work, but I'm not sure.

Ping @Major- @garyttierney

Major-:
@ryleykimmel also agree it should be in core

PR #376, opened 2017-12-31
[RFC] Port of Ruby combat framework to Kotlin

garyttierney:
While not in scope for the `kotlin-experiments` branch, this is something I'd still like to get some feedback on.  All of the basics are there for melee + ranged, including max hit / accuracy.  The DSL itself is almost a one-to-one port of what was done in Ruby, with some Kotlin specifics to make the framework more idiomatic.

See the Scimitar/Bow plugins for examples of how to configure new weapon classes and weapons.  `Scimitar` and `Bow` are `WeaponClass` objects that are defined by `melee_classes.kt` and `ranged_classes.kt`.


garyttierney:
My current TODO list for this (not counting the numerous `@todo`'s scattered around the code) can be found here: https://github.com/apollo-rsps/apollo/pull/376/files#diff-4553fbd36a1d88c8c2e744f5de0cc847

PR #384, opened 2018-03-27
Refactor embedded kotlin compiler out of plugin build system

garyttierney:
This is a large change that significantly simplifies the plugin build system by
having the kotlinc handle everything related to script compilation.  Since we
no longer use package name rewriting, the "packageName" option has been dropped
from the gradle script API and has been removed from existing plugins.

In addition to simplifying the build system, it should also significantly speed
up build times.  Previously we ran 2 build processes for every plugin and
didn't have fully working incremental compilation for scripts compared to 1
build process with built-in incremental compilation.

There were also some random pieces of code that referenced kotlinc directly, so
those have either been removed or refactored out.


game/plugin/skills/runecrafting/src/runecrafting.plugin.kts
```diff
@@ -2,7 +2,6 @@ import org.apollo.game.message.impl.*
 import org.apollo.game.model.entity.EquipmentConstants
 import org.apollo.game.model.event.impl.LoginEvent
 
-private val changeAltarObjectConfigId = 491
```
Major-: Probably should've done this change in a separate commit

garyttierney: oops, missed that

Major-:
LGTM

PR #400, opened 2018-07-18
Optimize collision matrix building

### LINKED ISSUE #328: Investigate sharp increase in memory usage

stevesoltys:
These changes reduce the startup time of Apollo by a few seconds ~~and set us up for saving a few more by adding the option to apply collision flags on-demand~~.

```
Jul 17, 2018 9:02:38 PM org.apollo.Server <init>
INFO: Starting Apollo...
Jul 17, 2018 9:02:38 PM org.apollo.Server init
INFO: Initialized Release 377.
Jul 17, 2018 9:02:39 PM org.apollo.ServiceManager startAll
INFO: Starting services...
Jul 17, 2018 9:02:47 PM org.apollo.Server bind
INFO: Ready for connections.
Jul 17, 2018 9:02:47 PM org.apollo.Server main
INFO: Starting apollo took 8719 ms.
```

Related to #328.

### DIFF
--- a/game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
+++ b/game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
@@ -1,19 +1,21 @@
 package org.apollo.game.model.area.collision;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import org.apollo.game.model.Direction;
 import org.apollo.game.model.Position;
 import org.apollo.game.model.area.Region;
+import org.apollo.game.model.area.RegionCoordinates;
 import org.apollo.game.model.area.RegionRepository;
 import org.apollo.game.model.area.collision.CollisionUpdate.DirectionFlag;
 import org.apollo.game.model.entity.EntityType;
 import org.apollo.game.model.entity.obj.GameObject;
 
 import java.util.Collection;
-import java.util.Comparator;
+import java.util.HashSet;
 import java.util.Map;
-import java.util.SortedSet;
-import java.util.TreeSet;
+import java.util.Set;
 
 import static org.apollo.game.model.entity.EntityType.DYNAMIC_OBJECT;
 import static org.apollo.game.model.entity.EntityType.STATIC_OBJECT;
@@ -25,20 +27,14 @@ import static org.apollo.game.model.entity.EntityType.STATIC_OBJECT;
 public final class CollisionManager {
 
 	/**
-	 * A comparator that sorts {@link Position}s by their X coordinate, then Y, then height.
+	 * A {@code HashMultimap} of region coordinates mapped to positions where the tile is completely blocked.
 	 */
-	private static final Comparator<Position> POSITION_COMPARATOR =
-		Comparator.comparingInt(Position::getX).thenComparingInt(Position::getY).thenComparingInt(Position::getHeight);
+	private final Multimap<RegionCoordinates, Position> blocked = HashMultimap.create();
 
 	/**
-	 * A {@code SortedSet} of positions where the tile is part of a bridged structure.
+	 * A {@code HashSet} of positions where the tile is part of a bridged structure.
 	 */
-	private final SortedSet<Position> bridges = new TreeSet<>(POSITION_COMPARATOR);
-
-	/**
-	 * A {@code SortedSet} of positions where the tile is completely blocked.
-	 */
-	private final SortedSet<Position> blocked = new TreeSet<>(POSITION_COMPARATOR);
+	private final Set<Position> bridges = new HashSet<>();
 
 	/**
 	 * The {@link RegionRepository} used to lookup {@link CollisionMatrix} objects.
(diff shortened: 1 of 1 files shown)


game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
```diff
@@ -25,20 +24,19 @@
...
-	private static final Comparator<Position> POSITION_COMPARATOR =
-		Comparator.comparingInt(Position::getX).thenComparingInt(Position::getY).thenComparingInt(Position::getHeight);
+	private final Multimap<RegionCoordinates, Position> blocked = HashMultimap.create();
 
 	/**
-	 * A {@code SortedSet} of positions where the tile is part of a bridged structure.
+	 * A {@code HashMultimap} of region coordinates mapped to positions where the tile is part of a bridged structure.
 	 */
-	private final SortedSet<Position> bridges = new TreeSet<>(POSITION_COMPARATOR);
+	private final Multimap<RegionCoordinates, Position> bridges = HashMultimap.create();
```
garyttierney: Why is this now per-region?

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
```diff
@@ -67,33 +65,55 @@ public void build(boolean rebuilding) {
...
-		for (Position tile : blocked) {
+		Collection<Position> regionTiles = blocked.get(buildRegion.getCoordinates());
+		Collection<Position> regionBridges = bridges.get(buildRegion.getCoordinates());
+
+		regionTiles.forEach(tile -> {
 			int x = tile.getX(), y = tile.getY();
 			int height = tile.getHeight();
 
-			if (bridges.contains(new Position(x, y, 1))) {
+			if (regionBridges != null && regionBridges.contains(new Position(x, y, 1))) {
```
garyttierney: What if `regionBridges` is null?

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
```diff
@@ -149,7 +173,7 @@ public void apply(CollisionUpdate update) {
 	 * {@code end} position using Bresenham's line algorithm.
 	 *
 	 * @param start The start position of the ray.
-	 * @param end The end position of the ray.
+	 * @param end   The end position of the ray.
```
garyttierney: Try and avoid whitespace/code style changes on code you haven't changed please.

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
```diff
@@ -111,7 +131,11 @@ public void apply(CollisionUpdate update) {
 			Position position = entry.getKey();
 
 			int height = position.getHeight();
-			if (bridges.contains(new Position(position.getX(), position.getY(), 1))) {
+
+			Position bridgePosition = new Position(position.getX(), position.getY(), 1);
+			Collection<Position> bridgePositions = bridges.get(bridgePosition.getRegionCoordinates());
+
+			if (bridgePositions != null && bridgePositions.contains(bridgePosition)) {
```
garyttierney: Same as `regionBridges` comment above.  I think there could be some edge cases around these conditions where an object extends into an adjacent region.  Any thoughts?

garyttierney (changes requested):

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
stevesoltys: I only eyeballed it, but making this per-region appeared to be slightly faster (the measure being startup time) than using a set of positions.

garyttierney: Faster than `TreeSet<Position>` or `HashSet<Position>`?

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
stevesoltys: I see what you're saying here. A solution could be to get the entities in the surrounding regions which extend into our region on line 113.

The caveat with that is that instead of ignoring flags, it will be applying flags for a region we haven't initialized yet. Maybe to fix that we could filter for positions that have the same region coordinates as the one we're applying flags to.

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
stevesoltys: If `regionBridges` is `null` then bridges doesn't contain the position, so BAU as per the prior functionality.

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
stevesoltys: `TreeSet<Position>`, I compared the original to this one.

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
stevesoltys: Actually I think I can remove these checks because `Multimap`returns an empty collection.

stevesoltys:
@garyttierney I've removed the on-demand collision flag functionality for now in favor of getting the optimizations merged in. The issue you brought up with adjacent regions is valid and the solution would bring us back up to ~15 seconds. We could re-introduce that functionality later when the on-demand region building is actually being used.

I also removed the whitespace changes as requested and made `bridges` a `HashSet`.

tlf30:
@stevesoltys does removing the flag not impact any currently implemented functionality? I would rather see slow load times on startup than functionality being disabled. 

Also, thank you for looking into optimizations.

stevesoltys:
@tlf30 nope, I was referring to changes I made previously being removed. This PR in its current state should not affect anything besides performance.

On-demand collision flag building can be added in a separate PR later on.

tlf30:
@stevesoltys sweet, thank you

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
```diff
@@ -25,20 +27,14 @@
 public final class CollisionManager {
 
 	/**
-	 * A comparator that sorts {@link Position}s by their X coordinate, then Y, then height.
+	 * A {@code HashMultimap} of region coordinates mapped to positions where the tile is completely blocked.
 	 */
-	private static final Comparator<Position> POSITION_COMPARATOR =
-		Comparator.comparingInt(Position::getX).thenComparingInt(Position::getY).thenComparingInt(Position::getHeight);
+	private final Multimap<RegionCoordinates, Position> blocked = HashMultimap.create();
```
garyttierney, line 32: Do we need the Map semantics here?  I think this can also be a HashSet.

garyttierney:
One more comment, LGTM besides that.

garyttierney (approved):

game/src/main/java/org/apollo/game/model/area/collision/CollisionManager.java
stevesoltys: I think this is faster because the other option would be looping over every tile in the region and doing a `HashSet#contains` before applying the tile as blocked. With the map we know which tiles are blocked already.

garyttierney: Er, right.  I missed the fact that we're actually iterating over these positions.

garyttierney:
Thanks, merged.

PR #401, opened 2018-07-22
Add bounds function for entities

stevesoltys:


game/src/main/java/org/apollo/game/model/entity/Entity.java
```diff
@@ -59,6 +61,13 @@ public World getWorld() {
 	 */
 	public abstract EntityType getEntityType();
 
+	/**
+	 * Returns the set of positions that this Entity occupies.
+	 *
+	 * @return The set of occupied positions.
+	 */
+	public abstract Set<Position> getBounds();
```
stevesoltys: Let me know if you'd like to change the name of this, if you can think of a better name. It might be confusing with `NPC#getBoundries()` already there.

rmcmk:
Maybe `getOccupiedPositions` or something along those lines as "bounds" isn't really fitting here. Also, should we be caching this sort of thing or will this change often enough for a caching mechanism to not be necessary?

stevesoltys:
Agreed @ryleykimmel, that might be better. As far as caching, I think it could make sense to cache the `GameObject` positions, assuming we will never change the position of a `GameObject` instance, but all other entities will change frequently from my understanding.

game/src/main/java/org/apollo/game/model/entity/Entity.java
garyttierney: I think it'd be better to return an object representing the rectangle of the bounds..  Recomputing every position in the `Mob`s bounds is going to result in a ton of temporary objects with a lot of NPCs active.  

A rectangle is also going to result in faster `contains` checks.

game/src/main/java/org/apollo/game/model/entity/Entity.java
```diff
@@ -34,6 +39,20 @@ public Entity(World world, Position position) {
...
 	public abstract boolean equals(Object obj);
 
+	/**
+	 * Gets the {@link EntityBounds} for this Entity.
+	 *
+	 * @return The EntityBounds.
+	 */
+	public EntityBounds getBounds() {
+
+		if(bounds == null) {
```
stevesoltys, line 49: There was a noticeable amount of overhead on startup without this. We also might not ever use this with certain entities like `Projectile`.

game/src/main/java/org/apollo/game/model/entity/EntityBounds.java
```diff
@@ -0,0 +1,93 @@
...
+	 *
+	 * @return The set of positions.
+	 */
+	public Position[] getPositions() {
+
+		if (entity.getPosition() == null) {
+			return (positions = new Position[0]);
+		}
+
+		if (!position.equals(entity.getPosition()) || width != entity.getWidth() || height != entity.getHeight()) {
```
stevesoltys: Is it better to just calculate these positions on each call since we have a `contains` function now, or should we keep it this way?

game/src/main/java/org/apollo/game/model/entity/GroundItem.java
```diff
@@ -80,6 +80,16 @@ public EntityType getEntityType() {
 		return EntityType.GROUND_ITEM;
 	}
 
+	@Override
+	public int getHeight() {
```
rmcmk: I think you mean `length`

game/src/main/java/org/apollo/game/model/entity/EntityBounds.java
```diff
@@ -0,0 +1,93 @@
...
+	 * <p>
+	 * If the width, height, or Position of the entity has changed since the last call, it will calculate a new
+	 * positions set before returning.
+	 *
+	 * @return The set of positions.
+	 */
+	public Position[] getPositions() {
+
+		if (entity.getPosition() == null) {
+			return (positions = new Position[0]);
```
rmcmk: Maybe make the default of `positions` an `EMPTY_POSITION_ARRAY`

game/src/main/java/org/apollo/game/model/entity/obj/GameObject.java
```diff
@@ -85,6 +86,23 @@ public int getType() {
 		return packed >> 2 & 0x3F;
 	}
 
+
+	@Override
+	public int getHeight() {
+		Direction direction = Direction.WNES[getOrientation()];
+
+		return direction == Direction.WEST || direction == Direction.EAST ?
```
rmcmk: This is used in a lot of places... (traversal stuff) maybe we could extract this functionality into its own method

game/src/main/java/org/apollo/game/model/entity/EntityBounds.java
```diff
@@ -0,0 +1,93 @@
...
+	 * Gets the set of positions that the Entity occupies.
+	 * <p>
+	 * If the width, height, or Position of the entity has changed since the last call, it will calculate a new
+	 * positions set before returning.
+	 *
+	 * @return The set of positions.
+	 */
+	public Position[] getPositions() {
+
+		if (entity.getPosition() == null) {
```
rmcmk: Is this actually ever `null`?

game/src/main/java/org/apollo/game/model/entity/obj/GameObject.java
stevesoltys: I think if we started using these new functions in the traversal stuff we wouldn't need another method, right? Assuming that code is getting the width / height of the object.

game/src/main/java/org/apollo/game/model/entity/EntityBounds.java
```diff
@@ -0,0 +1,89 @@
...
+	 * Gets the set of positions that the Entity occupies.
+	 * <p>
+	 * If the width, height, or Position of the entity has changed since the last call, it will calculate a new
+	 * positions set before returning.
+	 *
+	 * @return The set of positions.
+	 */
+	public Position[] getPositions() {
+
+		if (!position.equals(entity.getPosition()) || width != entity.getWidth() || height != entity.getLength()) {
```
stevesoltys: Is it better to just calculate these positions on each call since we have a contains function now, or should we keep it this way?

game/src/main/java/org/apollo/game/model/entity/obj/GameObject.java
rmcmk: Correct

game/src/main/java/org/apollo/game/model/entity/EntityBounds.java
```diff
@@ -0,0 +1,89 @@
...
+			position = entity.getPosition();
+			width = entity.getWidth();
+			height = entity.getLength();
+
+			positions = new Position[width * height];
+
+			for (int deltaX = 0; deltaX < width; deltaX++) {
+				for (int deltaY = 0; deltaY < height; deltaY++) {
+					positions[deltaX + deltaY] = position.step(deltaX, Direction.EAST).step(deltaY, Direction.NORTH);
```
rmcmk: Should we still be storing all of these individual positions? Wouldn't a Rectangle suffice? Maybe we should create some kind of Area data structure specifically for `Position`'s

game/src/main/java/org/apollo/game/model/entity/EntityBounds.java
rmcmk: I'm not sure if this is any better than your original PR, just looks decoupled...

rmcmk: I think I'm misunderstanding the reasoning for keeping all of the `Position`s handy rather than just keeping the area that we occupy

stevesoltys: Even if it isn't better than before, it's certainly not the same logic.. We also have `contains` now for checking if a position exists within the bounds, which satisfies the use case for the rectangle you're suggesting (it's really a polygon since width and length can be different).

This function is for getting the actual set of positions that the `Entity` occupies. The change is that it takes @garyttierney's suggestion of not creating a temporary object every call into account, but I'm not sure which is preferable now that we have `contains`.

If you're asking why this function exists, it's because I use it in a new distanced action implementation that I plan on merging later.

garyttierney: Agree with what Ryley said.  Where would the result of this call be used?

stevesoltys: If you prefer, we can leave this out, but it's possible we may need to merge something like it in later.

garyttierney: Let's leave it out for now.  I'm not convinced it's needed in the distanced action implementation, so can discuss that when we come to it.

rmcmk: I'm just trying to think of an application where this API would deem useful. If you've already written something or have an idea, do share. :)

rmcmk: Also, can't only `GameObject`'s be considered a polygon? Player/Npc/Projectile/GroundItem are always the same width and length, correct? Maybe the implementation should be altered only for `GameObject` if this is the case..

game/src/main/java/org/apollo/game/model/entity/EntityBounds.java
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * @param position The position.
+	 * @return A flag indicating whether or not the position exists within the Entity's bounds.
+	 */
+	public boolean contains(Position position) {
+		Position entityPosition = entity.getPosition();
+		int width = entity.getWidth();
+		int height = entity.getLength();
+
+		return position.getX() >= entityPosition.getX() && position.getX() < entityPosition.getX() + width &&
```
garyttierney: We should add local vars for `position.getX/Y()` and `entityPosition.getX/Y()` to make this a bit easier to read

garyttierney (changes requested):

game/src/main/java/org/apollo/game/model/entity/EntityBounds.java
stevesoltys: @ryleykimmel sure, I gotcha. Hop on slack when you get a chance, we had a small bit of discussion there.

garyttierney (approved):

game/src/main/java/org/apollo/game/model/entity/Entity.java
garyttierney: I think this a problem with how map loading is implemented, but we can tackle it later.

garyttierney:
Merged, thanks.

PR #402, opened 2018-07-22
Remove duplicate GameObject rotated checks

rmcmk:


garyttierney (approved):

stevesoltys:
@ryleykimmel @garyttierney I made a mistake, this should be the opposite:

```java
public boolean isRotated() {
    Direction orientation = Direction.WNES[getOrientation()];
    return orientation == Direction.NORTH || orientation == Direction.SOUTH;
}
```

Probably better to just swap the `getWidth()` and `getLength()` stuff in the width and length getters.

garyttierney:
@stevesoltys based on Ryley's testing, I think that will break collision maps.

cc @ryleykimmel this is the code you tested clipping against, right?

stevesoltys:
Yeah @garyttierney, I think this might break collision maps if merged. The code I merged before isn't breaking anything right now since it's not being used, but it is incorrect.

My code can be fixed in this PR once the above change is applied.

rmcmk:
It is the code I tested, are you sure it's incorrect? I was unable to find any issues.

cc @stevesoltys

stevesoltys:
Yes @ryleykimmel.

```java
if (orientation == 1 || orientation == 3) {	
    width = definition.getLength();	
    length = definition.getWidth();	
}
```

Length and width are reversed when the direction is east or west. We should be reversing them when the direction is north or south. You'll only see the issue in objects that aren't square.

Major-:
@stevesoltys Are you suggesting the code you pasted should actually be:
```
if (orientation == 0 || orientation == 2) {	
    width = definition.getLength();	
    length = definition.getWidth();	
}
```

Obviously need to correct this if so but otherwise this change looks good

stevesoltys:
No @Major-, the logic was correct before. There is a change in logic with this PR. Please re-review my comments and the code.

rmcmk:
There seems to be some confusion here. I think I understand. This code will break with diagonal object orientations. I missed the second call of `Direction.WNES` and `Direction.WNES_DIAGONAL`

This could be fixed by adding this functionality directly to `CollisionUpdate` or changing `GameObject#isRotated` to this:

```java
	public boolean isRotated() {
		int orientation = getOrientation();
		int type = getType();
		Direction direction = Direction.WNES[orientation];

		if (type == TRIANGULAR_CORNER.getValue() || type == RECTANGULAR_CORNER.getValue()) {
			direction = Direction.WNES_DIAGONAL[orientation];
		}

		return direction == Direction.NORTH || direction == Direction.SOUTH
			|| direction == Direction.NORTH_WEST || direction == Direction.SOUTH_EAST;
	}
```

stevesoltys:
Right @ryleykimmel. I was just pointing out that the `isRotated` orientations needed to be changed to:

```java
return direction == Direction.NORTH || direction == Direction.SOUTH;
```

but your change may solve additional diagonal issues as well.

PR #404, opened 2018-07-24
Add rotation functions to Direction

stevesoltys:


Major-:
Code LGTM, what's the use case for this? Things like doors?

garyttierney:
I think we should wait until this code has callers to merge this.

PR #405, opened 2018-07-26
Fix A* pathfinding when height is greater than zero

stevesoltys:


Major-:
LGTM although the import change isn't great; in apollo we don't use wildcard imports anywhere (in Java)

stevesoltys:
@Major- I've updated the pull request.

PR #408, opened 2018-07-31
Add custom bounds implementation for game objects

stevesoltys:
This is just a review request for now, unless the other contributors are interested in merging this before the remaining object types are supported. Apollo should be capable of supporting an updated `DistancedAction` once this functionality is available.

This PR in its current state is capable of getting the interaction positions for (AFAIK) all types of `GameObject` that are being used today in Apollo's plugins.

### DIFF
--- a/game/src/main/java/org/apollo/game/model/entity/obj/GameObjectBounds.java
+++ b/game/src/main/java/org/apollo/game/model/entity/obj/GameObjectBounds.java
@@ -1,0 +1,176 @@
+package org.apollo.game.model.entity.obj;
+
+import com.google.common.collect.ImmutableMap;
+import org.apollo.game.model.Direction;
+import org.apollo.game.model.Position;
+import org.apollo.game.model.entity.EntityBounds;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * The bounds of a {@link GameObject}.
+ *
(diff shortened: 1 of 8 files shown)


game/src/main/java/org/apollo/game/model/entity/obj/GameObjectBounds.java
```diff
@@ -0,0 +1,176 @@
...
+
+		return interactableDirections;
+	}
+
+	/**
+	 * Gets the set of positions where the GameObject can be interacted with.
+	 *
+	 * @return The set of interaction positions.
+	 */
+	private Set<Position> getInteractablePositions() {
```
garyttierney, line 136: I need an example of this in use.  I'm not sure that creating this `Set` all the time is optimal.  Maybe roll this up into your distanced action changes?

stevesoltys: @garyttierney In the distanced action we provide an `Entity` and check the interaction positions of that entity to decide when to trigger the action for a `Mob`.

Here's an example: https://gist.github.com/stevesoltys/197e7a604f8bfe495ae070c2db0c89c1

stevesoltys: Note that in my example, I've made the `getInteractionPositions` method abstract in `EntityBounds`.

PR #418, opened 2018-11-07
Make following and trading non-primary actions

lare96:
This PR makes "Follow" and "Trade with" appear **below** "Walk here" instead of being left click primary actions.


rmcmk:
You should fix this in the Kotlin version of this plugin. Ruby has been phased out and will be removed soon™

lare96:
Ahhh right, my mistake. Will do Ry!

lare96:
Done!

Major-:
@lare96 nice one, thanks

Issue #87, opened 2015-08-29
Running Depleting/Restoring Energy


WizardJesse1:
Shocked this never got done just making this a suggestion but I will write when I get time

WizardJesse1:
@Major- since It's very hard to get in touch with you I have this done but since in runescape running depleting is based on your agility level I do not know how I should do this since at the moment its done in the pulse of the walking queue.

lare96:
Run energy depletion is based on your weight, run energy restoration is based on your Agility level. You need to come up with an algorithm for both.

The one for restoration is found here http://2007.runescape.wikia.com/wiki/Energy

Major-:
@WizardJesse1 apollo doesn't actually support player weights yet, so you can just assume that the weight is 0 and leave a TODO. Each player has a [run_energy](https://github.com/apollo-rsps/apollo/blob/master/game/src/main/org/apollo/game/model/entity/Player.java#L68) attribute you can set/get the value of.

PR #423, opened 2019-04-25
Add Run Energy depleting/restoring

Tomm0017:
Addresses Issue https://github.com/apollo-rsps/apollo/issues/87 in which run energy depletion and restoring is requested

### DIFF
--- a/game/src/main/java/org/apollo/game/model/entity/Player.java
+++ b/game/src/main/java/org/apollo/game/model/entity/Player.java
@@ -473,8 +473,8 @@ public final class Player extends Mob {
 	 *
 	 * @return The run energy.
 	 */
-	public int getRunEnergy() {
-		Attribute<Integer> energy = attributes.get("run_energy");
+	public double getRunEnergy() {
+		Attribute<Double> energy = attributes.get("run_energy");
 		return energy.getValue();
 	}
 
--- a/game/src/main/java/org/apollo/game/model/entity/Player.java
+++ b/game/src/main/java/org/apollo/game/model/entity/Player.java
@@ -1059,9 +1059,10 @@ public final class Player extends Mob {
 	 * increment the run energy by.
 	 */
 	public void incrementRunEnergy(int agilityLevel) {
-		double increment = (8.0 + (agilityLevel / 6.0)) / 100.0;
-		double outcome = Math.min(MAXIMUM_RUN_ENERGY, getRunEnergy() + increment);
-		if (outcome != getRunEnergy()) {
+		double runEnergy = getRunEnergy();
+		if (runEnergy < MAXIMUM_RUN_ENERGY) {
+			double increment = (8.0 + (agilityLevel / 6.0)) / 100.0;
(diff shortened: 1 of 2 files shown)


game/src/main/java/org/apollo/game/sync/task/PrePlayerSynchronizationTask.java
```diff
@@ -58,6 +60,15 @@ public void run() {
 		Position old = player.getPosition();
 		player.getWalkingQueue().pulse();
 
+		if (player.getSecondDirection() != Direction.NONE) {
```
garyttierney, line 63: This isn't right. We should only decrement the run energy when the player actually moves (see `WalkingQueue`). The direction could be set but the collision maps said this is an invalid move.

Tomm0017: I wasn't fond of the condition, but it seemed to be the best fitting choice for what needed to be done. Sorry if I'm wrong here and just wasting your time, but I believe the direction is only set if the collision map allows the movement (see https://github.com/apollo-rsps/apollo/blob/kotlin-experiments/game/src/main/java/org/apollo/game/model/entity/WalkingQueue.java#L156). 

Also, am I wrong to assume that if I use ``WalkingQueue.isRunning()`` & ``WalkingQueue.size()``, _then_ I would run into the issue that you warned about? I believe when adding steps into the walking queue, it won't check for collision.

garyttierney: Hmm, I need to re-look at the assignment of `secondDirection`, you may be right there.

game/src/main/java/org/apollo/game/model/entity/Player.java
```diff
@@ -1035,4 +1040,30 @@ private void initSkills() {
 		skillSet.addListener(new LevelUpSkillListener(this));
 	}
 
+	/**
+	 * Decrement run energy.
+	 *
+	 * @param weight the amount of weight to be taken into account when
+	 * determining how much run energy will be depleted.
+	 */
+	public void decrementRunEnergy(double weight) {
```
garyttierney, line 1049: @Major- any thoughts on this being in core? Seems like it might be the responsibility of a plugin, but I'm not sure.

Major-: I think core is correct

game/src/main/java/org/apollo/game/sync/task/PrePlayerSynchronizationTask.java
Major-: @garyttierney any update on this?

garyttierney: The code is fine as is, and what we expect. That's the only assignment of `secondDirection` and the point @Tomm0017 made is valid.

PR #424, opened 2019-04-25
Add Essence Mining

Tomm0017:
Addresses Issue https://github.com/apollo-rsps/apollo/issues/184 where support for mining Rune and Pure essence is requested


Major-:
Looks great (sorry for the enormous delay in reviewing this), needs tests for the essence mining though.

PR #469, opened 2020-02-18
Heuristic abstraction modification

kiairatech:
Heuristics don't need to be an abstract class, they're an implementation of the estimation function across different heuristic types. No local variables are used either.


Major-:
Hi, thanks for your contribution. Partially for historical reasons, Apollo's Java codebase prefers to use abstract classes over interfaces as a defensive programming guideline, unless the type really needs to be an interface. I'm not necessarily opposed to changing this, but it'd need to address all the cases in one go, not just this one class.

If `Heuristic` being abstract is making it harder for downstream users to work with, I'm happy to open this again - let me know.

Issue #175, opened 2016-02-09
A PlayerSerializer that uses a database


Major-:
We want a `PlayerSerializer` that uses a database (postgres would be good, but making it implementation-independent would be nicer, if feasible). This is a must for good performance, and because the BinarySerializer is really just a hack left over from the original release.

Promises:
do you want a dedicated setting file?

garyttierney:
I think we do want some kind of centralized configuration file to hold all server settings.

Promises:
I have been doing some work on a postgre Serializer, it stores data in data/sql.xml, but I think Serializers should be plugins. That would make more sense for server owners that want forum integration.

Do you guys agree? or am I not seeing something?

garyttierney:
@Major- 

Thoughts on moving all data/*.xml files to one namespaced server.toml file?

@Promises 

I dislike this idea.  Plugins are meant for game logic, not arbitrary server extensions.  We already have a means for configuring serializers by putting an implementation on the classpath and editing the config file.  What does *forum integration* require that isn't solved by that?  Granted at the moment we have no way to pass serializer specific configuration to the implementation.  This is an issue which should be solved.

On the topic of your postgres serializer: how are you actually querying the database? i.e., writing SQL queries or an ORM/JPA.  I'm not sure what we're going to need here and we'll need to think about the schema to allow plugins to store attribute data.

Major-:
>Thoughts on moving all data/*.xml files to one namespaced server.toml file?

ya

Promises:
@garyttierney 
by following the documentation on postres site, I'm using pure SQL commands

garyttierney:
@Major- thoughts on that?  Do we want this Postgres specific or do we want a JDBC serializer?

Major-:
@Promises 
> by following the documentation on postres site, I'm using pure SQL commands

Don't forget to do this safely (i.e. beware SQL injection) as an attacker could craft e.g. a malicious username.

> Do we want this Postgres specific or do we want a JDBC serializer?

Hm. A db-agnostic implementation is nice for obvious reasons, but a postgres one would allow us to use postgres-specific features that aren't in the SQL standard, which might allow for better performance. Maybe the eventual goal would be both (assuming there _are_ useful non-standard features in postgres). 

Regardless of this I'm not sure if just writing raw SQL to a postgres connection is the way to go; using a driver is probably safer.

Promises:
[...]

PR #496, opened 2020-05-29
Player serialization from-and to SQL database

floatingpoint32:
Solving https://github.com/apollo-rsps/apollo/issues/175


game/build.gradle
```diff
@@ -30,6 +32,8 @@ dependencies {
     testImplementation group: 'org.powermock', name: 'powermock-module-junit4', version: powermockVersion
     testImplementation group: 'org.powermock', name: 'powermock-api-mockito2', version: powermockVersion
     testImplementation group: 'org.assertj', name: 'assertj-core', version: assertjVersion
+    testImplementation group: 'org.testcontainers', name: 'testcontainers', version: testContainersVersion
```
garyttierney: I don't like introducing Docker as a dependency to run tests and would prefer if the tests had some way to configure a JDBC URL. My own CI environment doesn't support `docker run`, but I do have other ways to start PostgreSQL databases servers.

game/src/main/java/org/apollo/game/account/Email.java
```diff
@@ -0,0 +1,41 @@
+package org.apollo.game.account;
+
+import java.util.Objects;
+
+/**
+ * An e-mail address.
+ * @author Sino
+ */
+public final class Email {
```
garyttierney: We can do without these value wrapper types (specifically `Email` and `PasswordHash`).

database/Dockerfile
```diff
@@ -0,0 +1,5 @@
+# Tell Docker to pull a Postgres 11 image running on Alpine
+FROM postgres:11-alpine
+
+# And then copy over the apollo.sql script into the container's entrypoint
+COPY apollo.sql /docker-entrypoint-initdb.d/init.sql
```
garyttierney: I don't want to start publishing custom PostgreSQL Docker images unless we need to compile native extensions. Instead of using the containers init scripts functionality we should be using something like Flyway to run migrations on the configured database.

database/start-db.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
garyttierney: Having the `docker-compose` configuration in here is fine, but we don't need shell scripts to run it. Users can just use the docker-compose commands directly.

game/src/main/java/org/apollo/game/database/ConnectionConfig.java
```diff
@@ -0,0 +1,130 @@
...
+
+import java.time.Duration;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Contains configurations specifically for (remote) database connections.
+ * @author Sino
+ */
+public final class ConnectionConfig {
```
garyttierney: All of this can be replaced by a JDBC URL. The PostgreSQL driver supports configuring timeouts (among other things) as driver properties in the JDBC URL.

game/src/main/java/org/apollo/game/database/ConnectionPool.java
```diff
@@ -0,0 +1,50 @@
...
+import java.sql.SQLException;
+
+/**
+ * A {@link ConnectionSupplier} that consumes and provides {@link Connection}s
+ * from a pool of reusable {@link Connection}s. When a connection is taken from
+ * the pool, it is up to the consumer to return it back to the pool to prevent
+ * leaking connections.
+ * @author Sino
+ */
+public final class ConnectionPool implements ConnectionSupplier {
```
garyttierney: We don't need this `DataSource` wrapper. Consumers who need a `Connection` can get it from the `DataSource` directly.

game/src/main/java/org/apollo/game/model/entity/Player.java
```diff
@@ -468,8 +468,8 @@ public PrivilegeLevel getPrivilegeLevel() {
 	 *
 	 * @return The run energy.
 	 */
-	public int getRunEnergy() {
-		Attribute<Integer> energy = attributes.get("run_energy");
+	public long getRunEnergy() {
```
garyttierney, line 471: Why have we changed the type of this attribute?

game/src/main/java/org/apollo/game/model/entity/Skill.java
```diff
@@ -129,6 +129,14 @@ public static String getName(int id) {
 		return SKILL_NAMES[id];
 	}
 
+	/**
+	 * Returns the amount of currently supported skill types.
+	 * @return The amount of currently supported skill types.
+	 */
+	public static int getCount() {
```
garyttierney: This method isn't an accessor, let's call it `count()` instead.

game/src/main/java/org/apollo/game/io/player/SlottedInventoryItem.java
```diff
@@ -0,0 +1,33 @@
...
+		return inventoryId;
+	}
+
+	SlottedItem getItem() {
+		return item;
+	}
+
+	@Override
+	public String toString() {
+		return "SlottedInventoryItem{inventoryId=" + inventoryId + ", item=" + item + "}";
```
garyttierney: We should use Guavas `MoreObjects` to generate `toString()` implementations like we have with other code.

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
```diff
@@ -1,33 +1,464 @@
...
+			.put(PrivilegeLevel.STANDARD, "player")
+			.put(PrivilegeLevel.MODERATOR, "moderator")
+			.put(PrivilegeLevel.ADMINISTRATOR, "administrator")
+			.build();
+
+	/**
+	 * Maps {@link AttributeType}s to the enum value that is known by the
+	 * database and back.
+	 */
+	private static final ImmutableBiMap<AttributeType, String> ATTRIBUTE_TYPE_ENUMS = ImmutableBiMap.<AttributeType, String>builder()
```
garyttierney, line 80: We target Java 11 now. We can use the `Map.of(...)` method instead.

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
```diff
@@ -1,33 +1,464 @@
...
+
+import static java.util.Objects.requireNonNull;
+
 /**
  * A {@link PlayerSerializer} that utilises {@code JDBC} to communicate with an SQL database containing player data.
  *
  * @author Major
+ * @author Sino
  */
+// TODO optimize use of collections to avoid boxing by using fastutil
```
garyttierney: Loading players isn't really a hot path. We don't want to use fastutil here.

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
```diff
@@ -1,33 +1,464 @@
...
+	private JdbcPlayerSerializer(World world, ConnectionSupplier connections) {
 		super(world);
+
+		this.connections = connections;
 	}
 
 	@Override
 	public void savePlayer(Player player) throws Exception {
-		throw new UnsupportedOperationException("JDBC saving is not supported at this time.");
+		try (Connection connection = connections.get()) {
```
garyttierney: This sequence of database queries needs to be wrapped in a transaction.

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
```diff
@@ -1,33 +1,464 @@
...
+
+			putInventory(connection, displayName, ITEM_BAG_ID, player.getInventory());
+			putInventory(connection, displayName, WORN_EQUIPMENT_ID, player.getEquipment());
+			putInventory(connection, displayName, BANK_ID, player.getBank());
+
+			putAttributes(connection, displayName, player.getAttributes());
+		}
+	}
+
+	private void putRank(Connection connection, Email email, PrivilegeLevel rank) throws SQLException {
```
garyttierney: I don't like the amount of manual interaction with `PreparedStatement`s and data mapping we have to do in here. I'd prefer if we used something like Spring's `JDBCTemplate` or jOOQ to handle splitting out the data mapping code (into spring `RowMapper`s or jOOQ `Converters`), common JDBC errors, and replace positional parameters with named parameters.

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
```diff
@@ -1,33 +1,464 @@
...
+			.put(AttributeType.BOOLEAN, "boolean")
+			.put(AttributeType.DOUBLE, "double")
+			.put(AttributeType.LONG, "long")
+			.build();
+
+	/**
+	 * The associated queries that are ran by this serializer.
+	 */
+	private static final String
+			GET_ACCOUNT_QUERY = "SELECT password_hash, rank FROM get_account(?::text)",
```
garyttierney: Why are these queries implemented as functions?

settings.gradle
```diff
@@ -3,6 +3,7 @@ import java.nio.file.Path
 
 rootProject.name = 'org.apollo'
 include ':cache'
+include ':database'
```
garyttierney: This isn't a Gradle module, so we shouldn't include it here.

database/README.md
```diff
@@ -0,0 +1,38 @@
+# Apollo - Database
```
garyttierney: This document is a bit information heavy for what could be a few lines in the README as a note "for local development a PostgreSQL database can be started using Docker and docker-compose, with the following commands: {...}".

The list of requirements and separate shell script is overkill. This information is already present on the Docker website (and is liable to change), and the script isn't comprehensive enough to substitute for a docker-compose cheat sheet (start, stop, logs, etc.) docker-compose commands will also work in Windows CMD, while the shell script won't.

database/apollo.sql
```diff
@@ -0,0 +1,506 @@
+CREATE EXTENSION citext;
```
garyttierney: Once we've got this schema into something that can manage it (i.e. Flyway) lets break it down and separate out schema primitives (like types and tables) from functions (whose migrations can be run repeatedly without the need to create new versions every time you update the function).

garyttierney (changes requested):

game/build.gradle
floatingpoint32: Removed dependency on testcontainers and updated test to take the JDBC URL as an environment variable.

game/src/main/java/org/apollo/game/account/Email.java
floatingpoint32: Removed the types.

database/start-db.sh
floatingpoint32: Removed the shell scripts.

game/src/main/java/org/apollo/game/database/ConnectionConfig.java
floatingpoint32: I've removed the entire database package and updated the serializer to take a `DataSource` instead.

game/src/main/java/org/apollo/game/model/entity/Skill.java
floatingpoint32: Applied.

game/src/main/java/org/apollo/game/io/player/SlottedInventoryItem.java
floatingpoint32: Applied.

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
floatingpoint32: Applied.

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
floatingpoint32: Removed comment.

game/src/main/java/org/apollo/game/model/entity/Player.java
floatingpoint32: I remember getting a `ClassCastException` whilst testing the `JdbcPlayerSerializer` so I updated it to reflect the available types in `AttributeType`, as there is no support for `Integer`. I don't know more than that

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
floatingpoint32: For now I require a `BiMap` implementation here to go back and forth with translating values. Do you perhaps know of a better way to do this? I don't think Guava supports a factory method like the default `Map` implementation does. Otherwise, this will have to wait until we've moved on to using Spring's `JDBCTemplate` or jOOQ

database/README.md
floatingpoint32: I agree. I've updated it. Sometimes less is better

game/src/main/java/org/apollo/game/io/player/JdbcPlayerSerializer.java
floatingpoint32: Because majority of them are slightly more complex than the usual one liners involving joins and such. Java doesn't support text blocks until 13 and we're on 11 now so that'd get a bit messy. Do you have a better idea?

garyttierney: If the only reason is to cut down on the size of string literals than we can store the queries in a kotlin file (which does support multi-line raw literals). The game module already has the Kotlin compiler enabled so this won't be an issue.

PR #515, opened 2020-09-13
Fix fishing rod tools bait

Arham4:
The Fishing Rod and Fly Fishing Rod's bait had to be flipped.

image


Arham4:
This is incorrect. Corrected in #516

